PR #1, opened 2018-07-25
Jroach/add repo skelton via generator

jeroach:


web-components/d2l-navigation-band.html
```diff
@@ -0,0 +1,27 @@
+<link rel="import" href="../bower_components/polymer/polymer.html">
+<link rel="import" href="../bower_components/d2l-colors/d2l-colors.html">
```
svanherk: This is missing from `bower.json`

README.md
```diff
@@ -1 +1,92 @@
-# navigation
\ No newline at end of file
+# navigation
```
svanherk: I think something may have gone weird here with the auto-filling (not sure how much of the readme the generator makes), but a lot of places that are referring to `d2l-navigation-band` should just be `d2l-navigation`.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  if [ $TRAVIS_PULL_REQUEST != false ] && [ $TRAVIS_SECURE_ENV_VARS == true ]; then
+    echo "Pull request with secure environment variables, running Sauce tests...";
+    npm run test:polymer:sauce || travis_terminate 1;
+  else
+    echo "Not a pull request and/or no secure environment variables, running headless tests...";
+    npm run test:polymer:local || travis_terminate 1;
+  fi
+env:
+  global:
+    SAUCE_USERNAME: Desire2Learn
```
svanherk: We need to add the Sauce password here in order for the sauce tests to run.  I can look into adding that, I have the instructions handy somewhere

web-components/d2l-navigation-band.html
```diff
@@ -0,0 +1,27 @@
+<link rel="import" href="../bower_components/polymer/polymer.html">
```
svanherk: This was the way bsi was setup, but the pattern I've seen for BrightspaceUI components is just to have the web components at the root (for example, https://github.com/BrightspaceUI/button)

README.md
svanherk: For example, lines 8, 14, 16, and the first occurrence in 26 should be just `d2l-navigation`

jeroach: Yeah the generator asks for a component name by default, so I gave it `d2l-navigation-band`. I then went through and updated most of the references to `navigation` since that was the root.

I was getting mixed up since we don't yet have a `d2l-navigation` wrapper component yet, but we will eventually. I likely should have used the generator with that name.

Anyway, ramble over - I'll fix it.

web-components/d2l-navigation-band.html
jeroach: Sounds good, I'll move it

web-components/d2l-navigation-band.html
jeroach: Good catch, will add.

.travis.yml
jeroach: Thanks!

CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+*		@jeroach
```
svanherk: This may work as is, but I've always seen the Code Owners file with everyone on one line for each file path.  Like https://github.com/Brightspace/d2l-custom-pages/blob/master/CODEOWNERS

package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "navigation",
+  "description": "Parent repo for all navigation based web components",
+  "private": true,
+  "scripts": {
+    "postinstall": "polymer install --variants",
+    "lint": "npm run lint:wc && npm run lint:html",
+    "lint:html": "eslint *.html demo/*.html test/*.html",
+    "lint:wc": "polymer lint -i d2l-navigation-band.html",
```
svanherk: Fine for now, but we may want to make this more general in case we forget to add new files here

test/index.html
```diff
@@ -0,0 +1,18 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
+		<title>d2l-navigation-band tests</title>
```
svanherk: Minor: Should be `d2l-navigation tests`

jeroach:
Thanks @svanherk - I just quickly addressed your concerns. Might as well do it while it's fresh 😄

PR #3, opened 2018-08-01
Jroach/add d2l navigation separator

jeroach:
US99013: https://rally1.rallydev.com/#/238990794704d/detail/userstory/241871087144


.gitignore
```diff
@@ -1,5 +1,2 @@
 bower_components
-bower_components-1.x
```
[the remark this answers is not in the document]
svanherk: We'll want to leave these, `package-lock.json` will be created (depending on your version of node I believe) but we don't want to include it, and the `1-x` stuff stops the Polymer 1 version of the bower components from being added here as well

d2l-navigation-separator.html
```diff
@@ -0,0 +1,28 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../d2l-icons/d2l-icons.html">
+
+<!--
+`d2l-navigation-separator`
+Polymer-based web component for a separator component to be used between buttons in a navigational element
+
+@demo demo/index.html
```
svanherk: This should be `demo/d2l-navigation-separator.html`, and we'll need to update the `navigation-band.html` to point to `demo/navigation-band.html`.  This tells the docs page where to link the "Demo" button to

jeroach: Good catch, will fix

.gitignore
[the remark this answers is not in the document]
jeroach: Thanks for letting me know - they have been added back.

jeroach:
Existing `divider.scss`
http://search.dev.d2l/source/xref/GitHub/Brightspace/brightspace-integration/sass/navigation/divider.scss

Also, as per discussion with Daves, we are going to let the parent component of the separator control when it resizes (the `@media` part of the linked file. Thus, it does not need to be included here.

d2l-navigation-separator.html
```diff
@@ -0,0 +1,32 @@
...
+
+@demo demo/d2l-navigation-separator.html
+-->
+
+<dom-module id="d2l-navigation-separator">
+	<template strip-whitespace>
+		<style>
+			span {
+				display: inline-block;
+				flex: 0 0 auto;
```
dlockhart: The `flex` I think should stay in BSI? The fact that the LE navigation is using flex layout to position the separator is somewhat specific to it IMO.

d2l-navigation-separator.html
```diff
@@ -0,0 +1,32 @@
...
+				margin: 0 9px;
+			}
+			d2l-icon {
+				color: var(--d2l-color-mica);
+			}
+		</style>
+		<span>
```
dlockhart: Is the `span` necessary anymore? i.e. could the CSS you're applying to the `span` just go on the host element?

```
:host {
    display: inline-block;
    margin: 0 9px;
}
```

jeroach: Does `:host` = parent? If so, then yes 😄  Just want to make sure I understand the syntax

dlockhart: `:host` would be the `d2l-navigation-separator` element itself.

jeroach: Ah - that should work too

jeroach: Working on the updates to all 3 PRs

jeroach: Done. Tested locally and looks correct 😄

d2l-navigation-separator.html
```diff
@@ -0,0 +1,29 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../d2l-icons/d2l-icons.html">
```
dlockhart: Nit: could you just import the parts of `d2l-icons` that you need? So specifically I think that would be `d2l-icon.html` and `tier2-icons.html`. It just lets the bundler optimize things a bit better by only bundling the things you actually need.

d2l-navigation-separator.html
```diff
@@ -0,0 +1,30 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../d2l-icons/d2l-icon.html">
+<link rel="import" href="../d2l-icons/tier2-icons.html">
+
+<!--
+`d2l-navigation-separator`
+Polymer-based web component for a separator component to be used between buttons in a navigational element
+
+@demo demo/d2l-navigation-separator.html
```
svanherk: Pretty sure this is my fault based on a previous comment.  This needs to be `@demo demo/navigation-separator.html` (without the `d2l`) to match the file name.

d2l-navigation-band.html
```diff
@@ -5,14 +5,14 @@
 `d2l-navigation-band`
 Polymer-based web component for a solid colour band that runs along the top of the navigational header
 
-@demo demo/index.html
+@demo demo/d2l-navigation-band.html
```
svanherk: Same here, this should be `@demo demo/navigation-band.html`

jeroach: Thanks! Fixed

PR #4, opened 2018-08-03
Jroach/add d2l navigation bar thick

jeroach:
Ready for initial review - Dave L suggested offline that some of this (e.g. `logo-area` should move back to BSI).


dlockhart:
Cool, let us know when you're back and it's in a state where you'd like us to have a look! Enjoy 🍹

jeroach:
Thanks Dave!

jeroach:
CI fail was due to Win 10 IE 11 tests timing out. All others passed. Kicking again.

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="d2l-navigation-shared-styles.html">
+
```
dlockhart: I'd suggest naming this guy `<d2l-navigation-header>`, which corresponds to what we call it in the LE. The reason being, one day the design may change and it may no longer be the thicker of the two bars... and then it's super confusing!

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
...
+				height: var(--d2l-navigation-header-height);
+			}
+
+			@media (max-width: 767px) {
+				.d2l-navigation-s-header-container {
+					height: var(--d2l-navigation-header-height-mobile);
+				}
+			}
+
+			div ::slotted(.d2l-navigation-s-header-open-button-wrapper) {
```
dlockhart: Let's remove this, since all the button stuff is going to be handled by the stuff Carol is doing.

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
...
+				display: inline-block;
+				flex: 0 1 auto;
+				height: 100%;
+			}
+
+			@media (min-width: 768px) {
+				div ::slotted(.d2l-navigation-s-header-open-button-wrapper) {
```
dlockhart: Same here, this can go.

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
...
+					display: none;
+				}
+			}
+
+			div ::slotted(.d2l-navigation-s-header-logo-area) {
```
dlockhart: Ditto -- the logo is an LE-navbar specific thing, so that CSS can continue to live in BSI. (until we see some more navigation use cases and decide to create a web component here for it).

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
...
+				}
+			}
+
+			div ::slotted(.d2l-navigation-s-header-actions) {
+				align-items: center;
+				display: flex;
+				flex: 0 0 auto;
+				height: 100%;
+			}
+		</style>
```
dlockhart: I think the `d2l-navigation-s-header` element can go away and its styles can just be applied to this element via `:host`.

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
...
+		<div class="d2l-navigation-s-header">
+			<div class="d2l-navigation-s-centerer">
+				<div class="d2l-navigation-s-gutters">
+					<div class="d2l-navigation-s-header-container">
+						<slot name="mobile-menu"></slot>
```
dlockhart: Let's remove the `mobile-menu` slot for now -- the LE can just put the mobile menu inside the `left` slot. And then when & if we want to support the concept of a menu in the future, we can pull that into here.

d2l-navigation-shared-styles.html
```diff
@@ -0,0 +1,43 @@
...
+				--d2l-navigation-header-height: 90px;
+				--d2l-navigation-header-height-mobile: 72px;
+
+				--d2l-navigation-margin-regular: 30px;
+
+				--d2l-page-padding-flex: 2.439%;
+				--d2l-page-padding-min: 15px;
+				--d2l-page-padding-max: 30px;
+			}
+
```
dlockhart: For all the CSS classes, let's go with `d2l-navigation-*`, dropping the `-s`. The `-s` was added a couple years ago when we moved the navbar rendering from client-side to server-side (`-s` for "server") rendering.

The add a postfix for each component in this repo. So it would be `d2l-navigation-centerer`, `d2l-navigation-header-container`, etc.

d2l-navigation-shared-styles.html
```diff
@@ -0,0 +1,43 @@
...
+		<style>
+			:host {
+				--d2l-navigation-header-height: 90px;
+				--d2l-navigation-header-height-mobile: 72px;
+
+				--d2l-navigation-margin-regular: 30px;
+
+				--d2l-page-padding-flex: 2.439%;
+				--d2l-page-padding-min: 15px;
+				--d2l-page-padding-max: 30px;
```
dlockhart: I'm a little torn about whether we should be exposing these variables. They'll be global, so if we do people could start using them and then we can't change them. They also don't provide a ton of value here (unlike in Sass) unless they end up getting used a lot.

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
...
+			        	<slot name="left"></slot>
+						<div class="d2l-navigation-s-gutter">
+							<slot name="middle"></slot>
```
dlockhart: Does anything actually need to go in the middle? I guess maybe not in the LE navbar but in some of the other designs?

d2l-navigation-bar-thick.html
```diff
@@ -0,0 +1,91 @@
...
+Polymer-based web component for the thick, upper section of the navigational element
+
+@demo demo/navigation-bar-thick.html
+-->
+
+<dom-module id="d2l-navigation-bar-thick">
+	<template strip-whitespace>
+		<style include="d2l-navigation-shared-styles">
+			.d2l-navigation-s-header {
+				border-bottom: 1px solid var(--d2l-color-gypsum);
```
dlockhart: @edwardg8 just made some fixes to BSI to solve some issues with the bottom-border. You'll want to make sure they get maintained here:
https://github.com/Brightspace/brightspace-integration/pull/949

edwardg8: When there are no links in the navbar, there'll be a border on `.d2l-navigation-s-header`

d2l-navigation-bar-thick.html
jeroach: Just the gutter (spacing). I doubt any elements. Main reason I was adding this was to allow it to define it's width (which in turn defines the position of the `right` component).

d2l-navigation-shared-styles.html
jeroach: Okay I'll remove the variables. This is the only usage so far. If we find another use case (or 2) then we can consider making them variables again.

d2l-navigation-bar-thick.html
jeroach: @edwardg8 Could you please link the matching LP/LE PR?

edwardg8: It's here https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/2528/overview

jeroach: thanks!

d2l-navigation-header.html
```diff
@@ -0,0 +1,71 @@
...
+`d2l-navigation-header`
+Polymer-based web component for the larger, upper section of the navigational element
+
+@demo demo/navigation-header.html
+-->
+
+<dom-module id="d2l-navigation-header">
+	<template strip-whitespace>
+		<style include="d2l-navigation-shared-styles">
+			:host {
```
svanherk: This and the section below (`.d2l-navigation-main-no-links `) seem to be the opposite of https://github.com/Brightspace/brightspace-integration/blob/master/sass/navigation/main.scss#L4.  Also, could we get rid of those from BSI in https://github.com/Brightspace/brightspace-integration/pull/921?

jeroach: @edwardg8 I think Stacey is right, but perhaps you could confirm. For the bottom border: what is the default colour, and what is the colour for the `no-links` case?

Is it:
Default: `$d2l-color-light-galena`
No links: `$d2l-color-gypsum`

jeroach: @svanherk Yes I think we can remove them - just did a code search. As long as we have the default and no link override specified on the header then all should be well.

edwardg8: `d2l-navigation-main-no-links` is intended. When the feature flag (link area color) is enabled, we want to use `$d2l-color-light-galena`, else use `$d2l-color-gypsum`. `$d2l-color-light-galena` is the new color added.

edwardg8: I think the existing code I had there missed the case when feature enabled, use  `$d2l-color-light-galena` for the border.

jeroach:
@svanherk @edwardg8 Okay, so I've removed the border css logic from the component, and thus left it in BSI for now.

After fiddling with it for quite some time, this seems to be the ideal situation.

Once we get to the parent `d2l-navigation` component, it will control the bottom border (as it wraps everything; regardless if the user is logged in or not).

jeroach:
@svanherk Any other concerns on this one?

PR #5, opened 2018-09-07
Cbusman/add d2l navigation button highlight

CarolBusman:
Creation of highlight buttons for the nav.


all-imports.html
```diff
@@ -1,2 +1,9 @@
 <link rel="import" href="./d2l-navigation-band.html">
```
svanherk: Seeing them all written out here, I wonder if we should just remove the "highlight" part in each name.  From that design call a few weeks ago, I got the sense we want things to be consistent.  The highlight part might change in the future, but in a navigation element we want people to use `navigation-button`s and `navigation-link`s.  @dlockhart @dbatiste Thoughts?

d2l-navigation-button-highlight-close.html
```diff
@@ -0,0 +1,47 @@
...
+This button contains an X icon.
+
+@demo demo/d2l-navigation-highlight.html d2l-navigation-button-highlight-close
+-->
+
+<dom-module id="d2l-navigation-button-highlight-close">
+    <template strip-whitespace>
+        <d2l-navigation-button-highlight-notification-icon
+            aria-expanded$="[[ariaExpanded]]"
+			aria-haspopup$="[[ariaHaspopup]]"
```
svanherk: There seems to be mismatched formatting/tabbing throughout all of the files.

d2l-button-behavior.html
```diff
@@ -0,0 +1,107 @@
+<link rel="import" href="../polymer/polymer.html">
```
svanherk: I think we should pull in `d2l-button` as a dependency, and then in each of the components just reference `<link rel="import" href="../d2l-button/d2l-button-behavior.html">` instead of `<link rel="import" href="./d2l-button-behavior.html">`.  Otherwise we need to try to keep these two in-sync, and they'll just be overwriting the same `window.D2L.PolymerBehaviors.Button` variable

d2l-navigation-button-highlight-notification-icon.html
```diff
@@ -0,0 +1,86 @@
...
+    <template strip-whitespace>
+        <style>
+            .highlight-button-icon {
+                color: var(--d2l-color-ferrite);
+            }
+            :host(:hover) .highlight-button-icon,
+            :host(:focus) .highlight-button-icon {
+                color: var(--d2l-color-celestine, #005694);
+            }
+            .highlight-button-notification-indicator {
```
svanherk: @dbatiste You'd mentioned that `my-courses` also needs to sue the notification dot, and we'd talked about splitting the logic out to a shared place. It seems like the styling is pretty light, what did you have in mind for that?  Was it @m6jones that was working on it?

d2l-navigation-link-back.html
```diff
@@ -0,0 +1,67 @@
...
+				href: {
+					type: String,
+					reflectToAttribute: true
+				},
+				_icon: {
+					type: String
+				}
+			},
+			attached: function() {
+				this._icon = this._getIcon(this.dir);
```
mtjandra: I don't see `dir` marked as a property. Is this intentional? Also shouldn't the direction be determined by the locale of the page rather than a attribute/property?

CarolBusman: The subtle button (https://github.com/BrightspaceUI/button/blob/master/d2l-button-subtle.html) changes its formatting based on dir="rtl", I thought that would be how the reading-direction would be passed?

d2l-navigation-button-highlight-close.html
```diff
@@ -0,0 +1,51 @@
...
+			formtarget$=[[formtarget]]
+			name$=[[name]]
+			alt-text=[[altText]]
+			type$=[[type]]
+			icon="d2l-tier1:close-large-thick">
+		</d2l-navigation-button-highlight-icon>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-button-highlight-close',
```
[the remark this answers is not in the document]
dbatiste: Is this a P1 vs P2 issue or a shadow-DOM vs shady-DOM?  The reason I ask is that shadow-DOM becomes the default for P2, whereas shady is the default for P1.  This sort of thing could crop up with shady vs. shadow-DOM because styles are scoped differently.

d2l-navigation-button-highlight-close.html
```diff
@@ -0,0 +1,51 @@
...
+			properties: {
+				altText: {
```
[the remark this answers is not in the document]
dbatiste: Good question.

If we do need a property, I think it should just be `text`.  My thinking is that alt is an image thing.  Aside, @dlockhart and I had a recent conversation about `text` vs `label`, and I think we came to agreement that `label` should only really be used on component APIs of form elements, like inputs.

d2l-navigation-link-back.html
```diff
@@ -0,0 +1,67 @@
...
+		<d2l-navigation-link-highlight href="[[href]]">
+			<d2l-icon icon="[[_icon]]"></d2l-icon>
+			<span>[[text]]</span>
+		</d2l-navigation-link-highlight>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-link-back',
+			properties: {
+				text: {
```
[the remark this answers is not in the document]
svanherk: Yeah I think we've seen that, and we can't hardcode in just the first part as that doesn't work in all translations.  Strings need to be translated all or nothing.  Maybe we want to have a default of "Back" though (and all its translations) and let people override it with the property?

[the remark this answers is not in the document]
mtjandra: Ok thanks yeah just double checking. In regards to different languages I have seen localization being applied from within the webcomponent if the text is never set by the user. They use the `d2l-localize-behavior` component.

d2l-navigation-button-highlight-close.html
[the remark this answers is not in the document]
CarolBusman: The problem seems to come from the style in the demo:
```
d2l-navigation-button-highlight-notification-icon,
d2l-navigation-button-highlight-image,
d2l-navigation-button-highlight-close {
	margin: 0px 5px;
	display: inline-block;
	height: 90px;
	/* border: black solid 1px; */
}
```
The `d2l-navigation-button-highlight-close` contains a `d2l-navigation-button-highlight-notification-icon`. 
In 1, it looks like its applying the margin to the close and the icon button inside of it, while in 2 it's only being applied to the close button.

all-imports.html
dbatiste: I think I would lean towards removing `highlight`.  It seems like there's no longer the need for the distinction.  Is that true?

d2l-navigation-button-highlight-close.html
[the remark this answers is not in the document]
CarolBusman: The `alt-text` eventually gets passed down to be set as the button's `title`. Rename this to `title`?

d2l-navigation-button-highlight.html
```diff
@@ -0,0 +1,84 @@
...
+				border-bottom-left-radius: 4px;
+				border-bottom-right-radius: 4px;
+				height: 4px;
+				left: -7px;
+				position: absolute;
+				top: 0;
+				width: calc(100% + 14px);
+			}
+			button,
+			button:focus {
```
dbatiste: Is `button:focus` needed here?

d2l-navigation-button-highlight-notification-icon.html
```diff
@@ -0,0 +1,86 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../d2l-colors/d2l-colors.html">
+<link rel="import" href="./d2l-navigation-button-highlight.html">
+<link rel="import" href="../d2l-icons/d2l-icons.html">
```
dbatiste: Not sure if it really matters, but `d2l-icons.html` includes a wack of stuff, but it looks like this component only needs `d2l-icon.html` and the tier 2 icon-set.  I think I am partial to listing them explicitly for sake of recognizing what actually needs to be imported.  

Also, in theory, if none of our components imports tier 3 for example, our bundle would exclude it, whereas if we use this `d2l-icons.html`, we'll be bundling all of it's imports regardless.

d2l-navigation-button-highlight-notification-icon.html
```diff
@@ -0,0 +1,86 @@
...
+				<d2l-icon icon="d2l-tier2:dot" class="highlight-button-notification-indicator"></d2l-icon>
+			</span>
+		</d2l-navigation-button-highlight>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-button-highlight-notification-icon',
+			properties: {
+				icon: {
+					type: String, reflectToAttribute: true
```
dbatiste: Nitpick: formatting.

d2l-navigation-button-highlight-notification-icon.html
```diff
@@ -0,0 +1,86 @@
...
+			.highlight-button-icon {
+				color: var(--d2l-color-ferrite);
+			}
+			:host(:hover) .highlight-button-icon,
+			:host(:focus) .highlight-button-icon {
+				color: var(--d2l-color-celestine, #005694);
+			}
+			.highlight-button-notification-indicator {
+				color: transparent;
+				right: calc(-50% + 4px);
```
dbatiste: Do we need RTL version of this?

d2l-navigation-button-highlight-notification-icon.html
CarolBusman: When I import just icon and tier 2, the icons used in the demo (`d2l-tier3:classes` and `d2l-tier3:notification-bell`) no longer show up.

d2l-navigation-button-highlight.html
```diff
@@ -0,0 +1,84 @@
...
+<dom-module id="d2l-navigation-button-highlight">
+	<template strip-whitespace>
+		<style>
+			.highlight-button-top-border {
+				background: transparent;
+				display: block;
+				border-bottom-left-radius: 4px;
+				border-bottom-right-radius: 4px;
+				height: 4px;
+				left: -7px;
```
dbatiste: Same here... do we need RTL version of this?

d2l-navigation-button-highlight-notification-icon.html
m6jones: Include d2l-tier3 if you require those icons as well.

d2l-navigation-button-highlight-notification-icon.html
CarolBusman: The notification orange circle is currently on the right of the button for both rtl and ltr languages in Spark.

svanherk: Hmm, I wonder if that was a design choice or just missed

d2l-navigation-button-highlight.html
CarolBusman: This element ends up being centered (it's moved left 7px, but its width is set to 100% + 14px, which should end up in a 7px overhang on each side).

d2l-navigation-link-back.html
dbatiste: It looks like this icon should automatically be mirrored, and should respect the dir set for the page.  (See icon def [here](https://github.com/BrightspaceUI/icons/blob/master/tier1-icons.html#L147).)  I think we should be able to just have `<d2l-icon icon="d2l-tier1:chevron-left"></d2l-icon>`, and so we should not need the `_icon` property, the logic in `attached`, or `_getIcon`.

Aside: a comprehensive dir check would be to get getComputedStyle like [this](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-focusable-arrowkeys-behavior.html#L128).  The best perf would simply be to use RTL selector in CSS and mirror the image, but while it would work for our app it's not 100% comprehensive either.  But, as mentioned, our icon mirroring should have us covered here.

d2l-navigation-link-highlight.html
```diff
@@ -0,0 +1,71 @@
...
+				position: relative;
+				text-decoration: none;
+				font-size: 14px;
+			}
+
+			a:hover .highlight-link-top-border,
+			a:focus .highlight-link-top-border {
+				background: var(--d2l-color-celestine, #005694);
+			}
+			:host {
```
dbatiste: A minor note: typically I place the `:host` select first when I have one.  Kind of a habit, and it tends to be where I (and probably others) look for it.

d2l-navigation-link-highlight.html
```diff
@@ -0,0 +1,71 @@
...
+				border-bottom-left-radius: 4px;
+				border-bottom-right-radius: 4px;
+				height: 4px;
+				left: -7px;
+				position: absolute;
+				top: 0;
+				width: calc(100% + 14px);
+			}
+			a,
+			a:focus {
```
dbatiste: I think it is better to have a separate `a:focus` selector with just the properties we want to change for that state.

d2l-navigation-link-highlight.html
```diff
@@ -0,0 +1,71 @@
...
+				cursor: pointer;
+				font-family: inherit;
+				height: 100%;
+				min-height: 40px;
+				margin: 0;
+				overflow: visible;
+				padding: 0;
+				position: relative;
+				text-decoration: none;
+				font-size: 14px;
```
dbatiste: All of our font-sizes should be in `rem`.  In this case `0.7rem`.

Will also likely need `outline: none;`.  Be sure to check browsers since I think they behave differently for this.

d2l-navigation-link-highlight.html
```diff
@@ -0,0 +1,71 @@
...
+			properties: {
+				href: {
+					type: String,
+					reflectToAttribute: true
+				}
+			},
+			attached: function() {
+				var anchor = Polymer.dom(this.root).querySelector('a');
+				if (anchor) {
+					anchor.style.lineHeight = (this.offsetHeight - 4) + 'px';
```
dbatiste: Curious about this - ideal we can avoid this measurement followed by mutation.  For a number of these, we'd get some DOM thrashing happening.

d2l-navigation-link-back.html
CarolBusman: Changing the icon back to `<d2l-icon icon="d2l-tier1:chevron-left"></d2l-icon>` and keeping the demo as `<d2l-navigation-link-back text="BACK." dir="rtl" href="https://www.example.org"></d2l-navigation-link-back>`, the chevron is still pointing to the left. Is there another step that needs to be added?

d2l-navigation-link-highlight.html
CarolBusman: The `img` in `d2l-navigation-link-image` was refusing to vertically center properly, so I added the `lineHeight` to give it something to center to.

d2l-navigation-button-highlight-close.html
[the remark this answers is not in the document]
dbatiste: So it sounds like shadow-DOM is the real cause then.  Since `d2l-navigation-button-highlight-notification-icon` is contained within the shadow-DOM of `d2l-navigation-button-highlight-close`, it is inside of a different DOM scope and will not be targeted by the `.d2l-navigation-button-highlight-notification-icon` CSS selector.  I suspect you'd see the same difference with P1 + shadow (by adding `?dom=shadow`) to query-string of page.

d2l-navigation-button-highlight-close.html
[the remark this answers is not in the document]
dbatiste: I think I would just stick with `text` to be consistent with our other button components.  ex. [d2l-button-icon](https://github.com/BrightspaceUI/button/blob/master/d2l-button-icon.html#L145)

d2l-navigation-button-highlight-notification-icon.html
dbatiste: Yeah I am not sure.  Might need Design to say.

d2l-navigation-button-highlight.html
dbatiste: I see.  👍

d2l-navigation-link-back.html
dbatiste: [This](https://github.com/PolymerElements/iron-iconset-svg/blob/master/iron-iconset-svg.html#L156) is the code that checks direction.  Note that it only gets run once (for a given icon-set), so timing would be important.

About a year ago, they added an optimization to this code `useGlobalRtlAttribute` since `getComputedStyle` can be expensive.  We currently don't set this attribute in our icon-sets, but perhaps could.  The default is `false`, so in our case I _think_ it's doing the `getComputedStyle` check so I would have thought your demo would work.  

I think would need to grab this branch and debug to say what exactly is happening in your demo.  This custom RTL code should not be needed.  :)

dbatiste: Suggestion - try hard-coding `dir="rtl"` on the body of your demo page and see if it makes a difference.  Obviously it would be important to remove this code, otherwise you'd be getting a flip on the opposite icon and it would seem like it's not working.

d2l-navigation-link-highlight.html
```diff
@@ -0,0 +1,71 @@
...
+			a:hover .highlight-link-top-border,
+			a:focus .highlight-link-top-border {
+				background: var(--d2l-color-celestine, #005694);
+			}
+			:host {
+				display: flex;
+				height: 100%;
+			}
+		</style>
+		<a href="[[href]]">
```
dbatiste: Similar to `D2L.PolymerBehaviors.Button.Behavior`, there is [D2L.PolymerBehaviors.Link.Behavior](https://github.com/BrightspaceUI/link/blob/master/d2l-link-behavior.html).  I wonder if we need to support `target` and `type`?  I think this would be important for actual navbar links, maybe not a big deal for these?

I suspect we also want `D2L.PolymerBehaviors.FocusableBehavior` and add the `d2l-focusable` class to the `a` so that if the `focus` method is called on `d2l-navigation-link-highlight`, focus will be directed to the internal anchor.

d2l-navigation-link-highlight.html
dbatiste: I think I would try to accomplish this in CSS.  I've used flex (with `align-items: center;`).  I think `d2l-button-icon` might use that approach.  Otherwise I may try applying equal padding/margin above and below to achieve vertical centering.

demo/d2l-navigation-highlight.html
```diff
@@ -0,0 +1,113 @@
...
+						} else {
+							child.style.border = 'black 1px solid';
+						}
+					}
+				}
+			}
+			function AssignClickEvent(className, handler) {
+				var elems = document.getElementsByClassName(className);
+				for (var i = 0; i < elems.length; i++) {
+					elems[i].addEventListener('click', handler, false);
```
dbatiste: Minor detail: `addEventListener` does not use event capturing by default, so the `false` is not needed here (and typically omitted).

d2l-navigation-link-back.html
CarolBusman: `dir="rtl"` in the `html` tag worked, but putting it in the `body` tag did not.
I guess remove the rtl example from the demo then, since it looks like I can't target one specific button.

m6jones: You will have to put it in the demo-snippet tag.

m6jones: The other option is to `<script>document.querySelector('body').setAttribute('dir','rtl');</script>` at the end of the demo file.

CarolBusman: Putting `dir="rtl"` on `demo-snippet` doesn't reverse the icon. :(

all-imports.html
dlockhart: I agree as well!

d2l-navigation-button-highlight-close.html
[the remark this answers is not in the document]
dlockhart: I agree with @mtjandra that I don't think the property is even needed, since it will always just be the text "close", no? So we could just include all the translations for "close" within this component and not need a property at all.

d2l-navigation-link-back.html
dlockhart: It's getting stripped from the `<body>` element because of [this bug](https://github.com/Polymer/tools/issues/654). It'll need to go on the `<html>` element or some other parent.

d2l-navigation-link-back.html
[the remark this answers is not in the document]
dlockhart: Yeah you could use a computed `text` property that when `text` is `null` returns `localize('text')`, otherwise returns the provided value.

bower.json
```diff
@@ -4,40 +4,42 @@
   "main": "index.html",
   "license": "Apache-2.0",
   "ignore": [
```
dlockhart: Nit: not sure why all the whitespace changes are happening here. @CarolBusman, do you have your IDE configured to use the `.editorconfig` file?

d2l-navigation-button-close.html
```diff
@@ -0,0 +1,51 @@
...
+			autofocus$=[[autofocus]]
+			form$=[[form]]
+			formaction$=[[formaction]]
+			formenctype$=[[formenctype]]
+			formmethod$=[[formmethod]]
+			formnovalidate$=[[formnovalidate]]
+			formtarget$=[[formtarget]]
+			name$=[[name]]
+			text=[[text]]
+			type$=[[type]]
```
dlockhart: I think all the attributes here except `text` and `icon` can be omitted? There should be no reason for people to disable a close button, or make it `aria-expanded`.

d2l-navigation-button-close.html
```diff
@@ -0,0 +1,51 @@
...
+		</d2l-navigation-button-icon>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-button-close',
+			properties: {
+				text: {
+					type: String,
+					reflectToAttribute: true
+				}
```
dlockhart: As suggested elsewhere, I think this can just always be the localized text "close". If down the road we need to make it customizable, we can always add a property in later.

d2l-navigation-button-image.html
```diff
@@ -0,0 +1,58 @@
...
+			form$=[[form]]
+			formaction$=[[formaction]]
+			formenctype$=[[formenctype]]
+			formmethod$=[[formmethod]]
+			formnovalidate$=[[formnovalidate]]
+			formtarget$=[[formtarget]]
+			name$=[[name]]
+			text=[[text]]
+			type$=[[type]]>
+			<img src="[[src]]">
```
dlockhart: Since this image is made redundant by the text, put `alt=""` to tell screen readers that they can safely ignore it. Otherwise, they'll read the full ugly `src` attribute.

d2l-navigation-button-image.html
```diff
@@ -0,0 +1,58 @@
...
+<link rel="import" href="../d2l-button/d2l-button-behavior.html">
+<link rel="import" href="../d2l-polymer-behaviors/d2l-focusable-behavior.html">
+
+<!--
+`d2l-navigation-button-image`
+Polymer-based web component for buttons used in the navigational header.
+This button contains a user-supplied image.
+
+@demo demo/d2l-navigation-button.html d2l-navigation-button-image
+-->
```
dlockhart: Remind me... what in the current nav does this web component correspond to?

d2l-navigation-button.html
```diff
@@ -0,0 +1,84 @@
...
+`d2l-navigation-button`
+Polymer-based web component for buttons used in the navigational header.
+
+@demo demo/d2l-navigation-button.html d2l-navigation-button
+-->
+
+<dom-module id="d2l-navigation-button">
+	<template strip-whitespace>
+		<style>
+			.button-top-border {
```
dlockhart: Nit: typically we prefix the full web component name in front of CSS classes -- so in this case, `d2l-navigation-button-top-border`. The reason is for when shadow DOM isn't available, all this CSS gets thrown into the global namespace, so the odds of your CSS class name colliding with one from something else on the page increase.

d2l-navigation-button.html
```diff
@@ -0,0 +1,84 @@
...
+				background: transparent;
+				display: block;
```
dlockhart: Another nit: CSS properties listed out alphabetically. It helps with finding a particular value, and also helps avoid conflicts if you're working on something with someone else.

d2l-navigation-link.html
```diff
@@ -0,0 +1,71 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../d2l-colors/d2l-colors.html">
+
+<!--
+`d2l-navigation-link`
+Polymer-based web component for buttons used in the navigational header.
+
+@demo demo/d2l-navigation-button.html d2l-navigation-link
```
dlockhart: Description and demo links still referring to "button"

d2l-navigation-link.html
```diff
@@ -0,0 +1,71 @@
...
+			}
+
+			a:hover .link-top-border,
+			a:focus .link-top-border {
+				background: var(--d2l-color-celestine, #005694);
+			}
+			:host {
+				display: flex;
+				height: 100%;
+			}
```
dlockhart: Seems like a fair bit of CSS could be shared here with the button version. I'd suggest one or more mixins in a shared style component.

d2l-navigation-link.html
```diff
@@ -0,0 +1,71 @@
...
+				href: {
+					type: String,
+					reflectToAttribute: true
+				}
+			},
+			attached: function() {
+				var anchor = Polymer.dom(this.root).querySelector('a');
+				if (anchor) {
+					anchor.style.lineHeight = (this.offsetHeight - 4) + 'px';
+				}
```
dlockhart: If this is a vertical positioning thing, I'd suggest flex positioning in CSS.

d2l-navigation-link.html
CarolBusman: They both point to the same demo. I included them together since they both look very similar to each other. Should I separate into a link demo and a button demo?

d2l-navigation-button-image.html
CarolBusman: I made a button module first before realizing that the logo was actually a link. I don't know if there's a current button in the nav that would use it.

d2l-navigation-link.html
dlockhart: Nah, all good since they're so related. I mostly just saw the "Polymer-based web component for buttons..." and then the component was called `d2l-navigation-link`.

d2l-navigation-button-image.html
dlockhart: Ok cool, then in that case I'd exclude it for now until we know for sure that we have a use case for it.

bower.json
CarolBusman: I made a change to VS Code's settings after these files were already created, and there were issues with mixed tabs and spaces. I think this file got caught in the whitespace fixes when I added a dependency.

d2l-navigation-button-highlight-notification-icon.html
m6jones: Yes, I was working on it. 
It is used here: https://github.com/BrightspaceHypermediaComponents/enrollments/blob/43f1a389434819391e8d5ee9828772513664801c/components/d2l-enrollment-card/d2l-enrollment-card.html#L157

CarolBusman:
I believe I've hit all the comments, this PR can be reviewed once more @dlockhart @dbatiste @m6jones @mtjandra

dbatiste:
LGTM!

d2l-navigation-button-close.html
```diff
@@ -0,0 +1,69 @@
...
+		<d2l-navigation-button-notification-icon
+			text=[[localize('Close')]]
+			icon="d2l-tier1:close-large-thick">
+		</d2l-navigation-button-icon>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-button-close',
+			properties: {
+				resources: {
```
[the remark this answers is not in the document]
dlockhart: I believe it needs to be for `localization-behavior` to work.

d2l-navigation-link-back.html
```diff
@@ -0,0 +1,82 @@
...
+Polymer-based web component for the back button used in the navigational header.
+
+@demo demo/d2l-navigation-button.html d2l-navigation-link-back
+-->
+
+<dom-module id="d2l-navigation-link-back">
+	<template strip-whitespace>
+		<style>
+			d2l-icon {
+				color: inherit;
```
dlockhart: We want a color other than ferrite for the back icon? I think this will inherit the body font color, which by default will be... ferrite.

d2l-navigation-link-back.html
```diff
@@ -0,0 +1,82 @@
...
+							'es': { 'Back': 'Volver' },
+							'fr': { 'Back': 'Précédent' },
+							'ja': { 'Back': '戻る' },
+							'ko': { 'Back': '뒤로' },
+							'nl': { 'Back': 'Terug' },
+							'pt': { 'Back': 'Voltar' },
+							'sv': { 'Back': 'Tillbaka' },
+							'tr': { 'Back': 'Geri' },
+							'zh': { 'Back': '返回' },
+							'zh-TW': { 'Back': '返回' }
```
dlockhart: Nit: normally we use lowercase keys for the language terms -- e.g. `back` instead of `Back`.

d2l-navigation-link-back.html
```diff
@@ -0,0 +1,82 @@
...
+				text: {
+					type: String,
+					value: null,
+					reflectToAttribute: true
+				},
+				href: {
+					type: String,
+					reflectToAttribute: true
+				},
+				resources: {
```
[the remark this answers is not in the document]
dlockhart: `localize-behavior` requires that `resources` be in properties.

d2l-navigation-link-back.html
CarolBusman: This is for the focus colour change, so that it picks up the change to celestine. Otherwise, the text was changing colour but the icon wasn't.

d2l-navigation-link-back.html
```diff
@@ -0,0 +1,82 @@
...
+			}
+			span {
+				padding: 0 5px;
+			}
+			span, d2l-icon {
+				vertical-align: middle;
+				height: 100%;
+			}
+		</style>
+		<d2l-navigation-link href="[[href]]" class="d2l-focusable">
```
dlockhart: Double-check that `href` is working... typically we need to use `href$="[[href]]"` to get it working properly.

d2l-navigation-button.html
```diff
@@ -0,0 +1,66 @@
...
+				@apply --d2l-navigation-highlight-border-hover-focus;
+			}
+			.d2l-navigation-button-top-border {
+				@apply --d2l-navigation-highlight-border;
+			}
+		</style>
+		<button
+			aria-expanded$="[[ariaExpanded]]"
+			aria-haspopup$="[[ariaHaspopup]]"
+			aria-label$="[[ariaLabel]]"
```
dlockhart: I wonder if we should just apply the `text` property to the `aria-label` of the button? In the existing buttons, we put an offscreen `<span>` containing the same text we put in the `title`, but an `aria-label` would accomplish the same thing.

d2l-navigation-link-back.html
dlockhart: Gotcha, thanks!

d2l-navigation-button-notification-icon.html
```diff
@@ -0,0 +1,83 @@
...
+			formenctype$=[[formenctype]]
+			formmethod$=[[formmethod]]
+			formnovalidate$=[[formnovalidate]]
+			formtarget$=[[formtarget]]
+			name$=[[name]]
+			text=[[text]]
+			type$=[[type]]>
+			<span class="d2l-navigation-button-icon-container">
+				<d2l-icon icon="[[icon]]" class="d2l-navigation-button-icon"></d2l-icon>
+				<d2l-icon icon="d2l-tier2:dot" class="d2l-navigation-button-notification-indicator"></d2l-icon>
```
dlockhart: In the existing navbar, the indicators have an offscreen label so that when an alert occurs, there's extra context in there for screen readers. For example, the message alert reads "Message alerts - You have new alerts".

I'd suggest exposing something similar here... like a new attribute called something like `notifcation-text` that gets put inside an offscreen label that's only visible when the notification dot is visible.

dlockhart: Once you add that offscreen text, you'll need to change the way you show/hide the notification dot. I'd recommend a CSS attribute selector:

    .d2l-navigation-button-notification-indicator {
        display: none;
    }
    :host([notification]) .d2l-navigation-button-notification-indicator {
        display: inline-block;
    }

d2l-navigation-link-back.html
CarolBusman: The links in the demo look like they're behaving properly. I can change it to `href$` just to be on the safe side though.

d2l-navigation-button-notification-icon.html
```diff
@@ -0,0 +1,83 @@
...
+					reflectToAttribute: true
+				},
+				notification: {
+					type: Boolean,
+					reflectToAttribute: true
+				},
+				text: {
+					type: String,
+					reflectToAttribute: true
+				}
```
dlockhart: Since they're not needed in attribute selectors, I think `text` and `icon` don't need `reflectToAttribute`.

d2l-navigation-link-back.html
dlockhart: I think we should. Not using the `$` mysteriously caused [crazy memory leaks](https://github.com/BrightspaceUI/icons/commit/2524990103520d9ea5a2773c809d6149dbc557bb) in other components that were using links.

dlockhart: Hmm, yeah I guess because it's not being applied to an `<a>` element it's all good! Keep it as-is.

d2l-navigation-link-image.html
```diff
@@ -0,0 +1,44 @@
...
+			is: 'd2l-navigation-link-image',
+			properties: {
+				src: {
+					type: String,
+					reflectToAttribute: true
+				},
+				href: {
+					type: String,
+					reflectToAttribute: true
+				}
```
dlockhart: Same thing here -- `reflectToAttribute` shouldn't be needed.

d2l-navigation-link-image.html
```diff
@@ -0,0 +1,44 @@
...
+			}
+			.d2l-navigation-link-image-container {
+				height: 100%;
+				vertical-align: middle;
+				align-items: center;
+				display: inline-flex;
+			}
+		</style>
+		<d2l-navigation-link href="[[href]]">
+			<span class="d2l-navigation-link-image-container"><img src="[[src]]"></span>
```
dlockhart: We need some kind of accessible label/tooltip for this. I think the component itself should have a `text` property (required), that we then apply to the `[[text]]` and possible also put it as a title on the `<a>` tag under `<d2l-navigation-link>`. This would match how the current navbar logo works.

d2l-navigation-link.html
```diff
@@ -0,0 +1,56 @@
...
+			}
+			a:hover .d2l-navigation-link-top-border,
+			a:focus .d2l-navigation-link-top-border {
+				@apply --d2l-navigation-highlight-border-hover-focus;
+			}
+			.d2l-navigation-link-top-border {
+				@apply --d2l-navigation-highlight-border;
+			}
+		</style>
+		<a class="d2l-focusable" href="[[href]]">
```
dlockhart: This one needs `$`. I can't explain why, but there was a [crazy memory leak](https://github.com/BrightspaceUI/icons/commit/2524990103520d9ea5a2773c809d6149dbc557bb) caused by this in the past.

mtjandra: Do you know if you experience this with all `href` attributes, or specifically for the `<a>` tag?

dlockhart: I believe it was just on the `<a>` tag, and specific to IE11 (surprise, surprise).

d2l-navigation-link.html
```diff
@@ -0,0 +1,56 @@
...
+			<slot></slot>
+		</a>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-link',
+			properties: {
+				href: {
+					type: String,
+					reflectToAttribute: true
```
dlockhart: Not needed.

d2l-navigation-link.html
```diff
@@ -0,0 +1,56 @@
...
+			}
+			a:hover .d2l-navigation-link-top-border,
+			a:focus .d2l-navigation-link-top-border {
+				@apply --d2l-navigation-highlight-border-hover-focus;
+			}
+			.d2l-navigation-link-top-border {
+				@apply --d2l-navigation-highlight-border;
+			}
+		</style>
+		<a class="d2l-focusable" href$="[[href]]">
```
dlockhart: I think we should transfer the `title` attribute here, so that consumers can set the tooltip.

test/d2l-navigation-button.html
```diff
@@ -0,0 +1,268 @@
...
+				setup(function() {
+					button = fixture('closeButton');
+				});
+				testInstantiation('d2l-navigation-button-close');
+				test('should pass text down to the base button', function() {
+					var title = button.$$('d2l-navigation-button-notification-icon').text;
+					assert.equal('Close', title);
+				});
+
+				translations.close.forEach(function(t) {
```
dlockhart: These tests seem like a little overkill, since really you're testing the `localize-behavior` at this point. I'd suggest just testing that it's translated to English properly and leave the specifics for other languages to the behaviour.

CarolBusman:
@dlockhart I believe I've touched all the comments, let me know if I've missed any, or if there are any other issues with this PR.

PR #7, opened 2018-09-21
name in bower.json should match registered name

dlockhart:
This is so you can go `bower link d2l-navigation` and have it actually work. :)


svanherk:
Thanks @dlockhart!

PR #8, opened 2018-09-21
treating header host as a block element, removing need for a <div>

dlockhart:
Just removing a redundant element. @jeroach, by default `:host` elements are `inline-block`, so if you want them to be anything other than that you need to specify it. In this case, we want the header to be a block element like a `<div>`.


jeroach:
Thanks for letting me know Dave!

PR #9, opened 2018-10-03
Jroach/d2l navigation bottom bar

jeroach:
As per US99018: https://rally1.rallydev.com/#/238990794704d/detail/userstory/241874923132

BSI PR: https://github.com/Brightspace/brightspace-integration/pull/1021/
platformTools PR: https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/2634/overview

FYI - I plan to add more unit tests. But I don't want to spend a ton of time there if changes are required to the base component. More tests will happen in a follow up PR.


demo/navigation-bottom-bar.html
```diff
@@ -0,0 +1,40 @@
...
+			}
+		</style>
+	</head>
+	<body unresolved class="d2l-typography">
+		<div class="vertical-section-container centered">
+			<h3>d2l-navigation-bottom-bar</h3>
+			<demo-snippet>
+				<template>
+					<d2l-navigation-bottom-bar>
+						<div slot="main" class="d2l-navigation-s-main-wrapper" has-edit-menu>Has edit menu</div>
```
mtjandra: Suggestion: Colour the nav bar so we can see what the different between the two is. Seems the only difference is we have a right margin when `has-edit-menu` is true.

Is we planning to add a slot on the right to allow for a menu button in future, or are we planning to add the menu box button in a future PR?

jeroach: Thanks Miles! Re: the elements increasing/decreasing and the bar disappearing, this is what is currently supported. However, it doesn't make sense to add it to the demo since the media queries that control hiding and showing live with the parent web component. This was a decision made after consulting with the Daves back when we encountered the same issue for the `d2l-navigation-header` web component.

I will work on colouring the navbar to show the variation in the padding when `has-edit-menu` is set.

Also, as per the AC of the story, there will only be the single slot (as is now) that spans the entire width of the component. The code using the component can decide what to show within.

d2l-navigation-bottom-bar.html
```diff
@@ -0,0 +1,89 @@
...
+					<slot name="main"></slot>
+				</div>
+			</div>
+		</div>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-bottom-bar',
+			properties: {
+				hasbranding: {
```
dlockhart: Which feature flag is controlling this? We might be at a point in its lifecycle that we can remove it from code, so that could simplify things here. @edwardg8 FYI

dlockhart: Looks like it's `daylight-navbar-link-area-color`, which was enabled in `10.8.3`. Four releases is probably enough for this one I think -- I'd say it's safe to remove from code, which would mean you don't need to worry about it here.

d2l-navigation-bottom-bar.html
```diff
@@ -0,0 +1,89 @@
...
+`d2l-navigation-bottom-bar`
+Polymer-based web component for the smaller, lower section of the navigational element
+
+@demo demo/navigation-bottom-bar.html
+-->
+
+<dom-module id="d2l-navigation-bottom-bar">
+	<template strip-whitespace>
+		<style include="d2l-navigation-shared-styles">
+			.d2l-navigation-s-main {
```
dlockhart: I might recommend dropping the `-s` from the class name here. Two reasons:
- By changing it from what it was, you might catch places that were accidentally depending on CSS being there or the old class being present
- The `-s` was originally added to distinguish our original client-side navbar from the one that's used now which is server-side rendered. `s` for "server-side", which this is definitely not. :)

d2l-navigation-bottom-bar.html
```diff
@@ -0,0 +1,89 @@
...
+			.d2l-navigation-main-tb .d2l-navigation-s-edit-menu {
+				top: 12.5px;
+			}
+
+			.d2l-navigation-main-tb[main-tb-hover] .d2l-navigation-s-edit-menu,
+			.d2l-navigation-main-tb[main-tb-focus] .d2l-navigation-s-edit-menu {
+				opacity: 1;
+			}
+
+			div ::slotted(.d2l-navigation-s-main-wrapper) {
```
dlockhart: There's a fair bit of mixed knowledge here... like the web component is aware of this child that can have a class name of `d2l-navigation-s-main-wrapper`, and it's also aware that the child can have these `data-more`, `data-loading` and `data-unmeasured` attributes on it.

A couple options:
1) Keep the CSS for `.d2l-navigation-s-main-wrapper` and all the `data-` stuff in BSI
2) Move everything into here

I don't think we have enough information about how others will use navigation to know whether 2) is necessary, so I'd lean towards 1) I think.

d2l-navigation-bottom-bar.html
dlockhart: It should also likely match the web component itself? i.e. `d2l-navigation-bottom-bar`?

d2l-navigation-bottom-bar.html
jeroach: Good call - I'll dig a bit more, but I think you're right in that option 1) makes more sense.

svanherk: Yeah I agree with 1) as well.  I imagined this component being a lot simpler, and just being able to accept the LMS-specific branding changes but not have knowledge of them, and jsut be the default we expect for immersive nav.  I talked to Jeff G and there's currently no plans for the immersive nav to use these special branding settings or have that carry over throughout the LMS, so I think it makes sense to leave those in BSI.

d2l-navigation-bottom-bar.html
```diff
@@ -0,0 +1,42 @@
...
+<dom-module id="d2l-navigation-bottom-bar">
+	<template strip-whitespace>
+		<style include="d2l-navigation-shared-styles">
+			:host {
+				border-top: 1px solid var(--d2l-color-gypsum);
+				border-bottom: 1px solid var(--d2l-color-gypsum);
+			}
+
+			.d2l-navigation-main-tb .d2l-navigation-s-edit-menu {
+				top: 12.5px;
```
dbatiste: Probably makes sense to pick an integer values if specifying `px` units.  I'm not sure all browsers will treat this value the same - one may truncate, another may round up, etc.

jeroach:
@mtjandra @dlockhart @dbatiste All comments should be addressed between this PR, and the linked PRs.

Only odd thing I've noticed is the `@media` query to hide the bottom nav on mobile screens does not work with `?dom=shadow`. Is this expected? If not, I'll look into fixing. Thanks all!

mtjandra:
Height of the bottom-bar seems to adjust based on its contents. Unsure if this is intentional or not.

image

dlockhart:
I'd personally stay away from Galen for now if you can. It's pretty heavy, slow and can be glitchy. I'd much rather see us take the time to find a more suitable perceptual diff tool that can hook into our CI systems. I've got a shortlist to investigate if anyone's interested!

PR #11, opened 2018-10-11
Moving display styling to component so they are inline-block by default.

CarolBusman:


CarolBusman:
@mtjandra The current styling for the buttons has them as `flex` but these components need to be `inline-block` in order to display properly in the navigation bar. I was planning on adding a class with the correct display in BSI but after discussing with @svanherk, thought it would make more sense to add it as a default in the component itself.

svanherk:
I'd hold off on the effort of adding galen tests for now... we're trying to figure out how well that fits with the upcoming framework migration, and there's some effort being put into a perpetual diff tool

PR #12, opened 2018-10-11
Initial CSS style updates to fix long navbar

mtjandra:
* Getting started
* Initial changes seem to look okay in chrome

In original I see references to 
```
-ms-flex-align: center;
-ms-flexbox
-ms-flex: 0 1 auto;
```

Might need some ie/edge specific styling

### DIFF
--- a/d2l-navigation-header.html
+++ b/d2l-navigation-header.html
@@ -30,16 +30,22 @@ Polymer-based web component for the larger, upper section of the navigational el
 			}
 
 			div ::slotted(.d2l-navigation-header-left) {
+				-ms-flex-align: center;
 				align-items: center;
+				display: -ms-flexbox;
 				display: flex;
+				-ms-flex: 0 1 auto;
 				flex: 0 1 auto;
 				height: 100%;
 				overflow: hidden;
 			}
 
 			div ::slotted(.d2l-navigation-header-right) {
+				-ms-flex-align: center;
 				align-items: center;
+				display: -ms-flexbox;
 				display: flex;
+				-ms-flex: 0 0 auto;
 				flex: 0 0 auto;
 				height: 100%;
 			}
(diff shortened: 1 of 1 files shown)


d2l-navigation-header.html
```diff
@@ -1,18 +1,20 @@
...
+
+@demo demo/navigation-header.html
+-->
+
+<dom-module id="d2l-navigation-header">
+	<template strip-whitespace>
 		<style include="d2l-navigation-shared-styles">
 			:host {
 				display: block;
+				margin: 0 auto;
```
svanherk: We are applying this style over here: https://github.com/BrightspaceUI/navigation/blob/master/d2l-navigation-shared-styles.html#L10
Is this actually broken, or just refactoring? The reason we have it in the shared styles is because the `bottom-bar` uses it too

mtjandra: I found it wasn't working with that, and needed to set it in the host. I'm not too great at styling so I'm sure there are things to improve. Right now it isn't working in IE, playing around with it. Below you see overflow in IE

image

jeroach:
Thanks Miles!! If the shared style isn't working, we should likely put it in the host of the `bottom-bar` as well and remove it from the shared styles. Not important for the cert fix, but just wanted to note it.

Please let me know if I can help!

svanherk:
Update (reminded by Josh's comments): I'm not seeing any issues with the shared styles in any of the browsers

jeroach:
@svanherk As in the shared style works as expected for `navigation-header` and that it does NOT need to be explicitly put in the `:host` element?

svanherk:
> @svanherk As in the shared style works as expected for navigation-header and that it does NOT need to be explicitly put in the :host element?

Yes.  This PR can be closed, if we're good with the ones I put up.

jeroach:
Awesome! Thanks! Sounds good to me. I'll close it.

PR #14, opened 2018-10-15
DE31707 - Adjustments as part of the long course names fix

svanherk:
- Fix the flex values
- Make the demo more useful

Most of the changes for the fix will be in the BSI PR, not this component.  Will put up that PR next.

### DIFF
--- a/d2l-navigation-header.html
+++ b/d2l-navigation-header.html
@@ -31,8 +31,15 @@ Polymer-based web component for the larger, upper section of the navigational el
 			div ::slotted(.d2l-navigation-header-right) {
 				align-items: center;
 				display: flex;
-				flex: 0 0 auto;
 				height: 100%;
+			}
+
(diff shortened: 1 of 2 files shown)


d2l-navigation-header.html
```diff
@@ -31,10 +31,17 @@
...
 				display: flex;
-				flex: 0 0 auto;
 				height: 100%;
 			}
 
+			div ::slotted(.d2l-navigation-header-left) {
+				flex: 0 1 auto;
+			}
+
+			div ::slotted(.d2l-navigation-header-right) {
```
mtjandra: Why did you split this out? There was already a section fo `div ::slotted(.d2l-navigation-header-right)`

svanherk: The section above is for both left and right, so I just wanted to make it really obvious which `flex` setting goes with which direction, rather than setting one for both and then overriding for one.

PR #20, opened 2018-10-22
Changing default button font to inherit instead of setting manually.

CarolBusman:


d2l-navigation-highlight-styles.html
```diff
@@ -11,7 +11,8 @@
 					border: transparent;
 					color: var(--d2l-color-ferrite);
 					cursor: pointer;
-					font-size: 0.875rem;
+					font-family: inherit;
```
svanherk: This used to be broken and we had to work around it (for example, https://github.com/BrightspaceUI/inputs/blob/master/d2l-input-shared-styles.html#L32) but it looks like the issue has since been fixed: https://github.com/webcomponents/shadycss/issues/91

Are you seeing the latest version of `shadycss` in your `bower_components`?

CarolBusman: I thought so? On further examination however, looks like it's working in my 1.X demo, but not the mainline. I'll move the styling to the component itself instead of using `@apply`.

PR #22, opened 2018-10-26
Add d2l-navigation and sample demo

jeroach:


d2l-navigation.html
```diff
@@ -0,0 +1,49 @@
...
+			}
+			.d2l-navigation-shadow-gradient {
+				background: linear-gradient(to bottom,  rgba(249,250,251,1) 0%,rgba(249,250,251,0) 100%);
+				bottom: -150px;
+				height: 150px;
+				width: 100%;
+				pointer-events: none;
+				position: absolute;
+				z-index: -100;
+			}
```
mtjandra: Nice, looks to match styling in BSI

d2l-navigation.html
```diff
@@ -0,0 +1,49 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="d2l-navigation-band.html">
+<link rel="import" href="d2l-navigation-header.html">
```
svanherk: A lot of these aren't used here, but it raises an interesting point - I imagined that these _would_ be included in the final "whole nav component".  _But_, then you'd be duplicating the shadow style for the immersive nav.  So maybe what we want is one more web component on top of this one?

So let's say we renamed this to `d2l-navigation-container`.  Do we want a `d2l-navigation-main` that looks like this:
```
<d2l-navigation-container>
  <d2l-navigation-header>
    <slot name="main-left"><div slot="left"><\div><\slot>
    <slot name="main-right"><div slot="right"><\div><\slot>
  <\d2l-navigation-header>
  <d2l-navigation-bottom-bar>
    <slot name="main-bottom"><\slot>
  <\d2l-navigation-bottom-bar>
<\d2l-navigation-container>
```

And then `d2l-immersive-nav` is:
```
<d2l-navigation-container>
  <d2l-navigation-header>
    <slot name="immersive-left"><div slot="left"><\div><\slot>
    <slot name="immersive-right"><div slot="right"><\div><\slot>
  <\d2l-navigation-header>
<\d2l-navigation-container>
```

(Not sure if the slots are setup correctly, this is just a rough idea)

test/navigation.html
```diff
@@ -0,0 +1,26 @@
...
+		<test-fixture id="basic">
+			<template>
+				<d2l-navigation></d2l-navigation>
+			</template>
+		</test-fixture>
+		<script>
+			suite('d2l-navigation', function() {
+				test('instantiating the element works', function() {
+					var nav = fixture('basic');
+					assert.equal(nav.is, 'd2l-navigation');
```
mtjandra: Could maybe check to ensure the following div/class/styling is rendered:

```
<div class="d2l-navigation-shadow-drop-border"></div>
<div class="d2l-navigation-shadow-gradient"></div>
```

May need to give them ids

d2l-navigation.html
mtjandra: Hey sorry I probably missed an earlier discussion but why are we making this container. Is it just for the shadow? Also I thought the immersive nav is different from the "thick nav", so why re-using d2l-navigation-header?

Feel free to send pictures if I'm getting things confused

svanherk: The immersive nav only has one layer, while the main one has two.  We originally thought we'd want to use the thinner component for immersive, but we realized in an offline convo we probably want the header because of the left and right slots.  Once Jeff gets back to us with heights, we can make the height configurable and set the heights differently for the main nav versus the immersive one. 

So yes, this container would become just for the shadow, and it could control the width of the navbar/page if we want (that will also need to depend on Jeff's design.

mtjandra: Ah okay I see so container for
* Shadow
* Height
* Width

The only thing that throws me off with your example is nothing stops you from something like
```
<slot name="immersive-left"><div slot="left"><\div><\slot>
<slot name="main-right"><div slot="right"><\div><\slot>
```

Also you guys probably know better than me, but is two slots the way we want to move forward with? Given that overflow highlighting  issue we had previously?

svanherk: This would be in the component, so people can't switch them like that.  We're picking the names they have to use, and if they use a slot name that doesn't exist, it just won't work.

In my idea it's creating two new slots, but _all_ they do is expose the slots of `d2l-navigation-header`.  I started questioning whether this was a good idea though, and had a chat with @dbatiste about it.  In the end, it takes flexibility away from the user, which can be good (if we want consistency) and bad (if we want flexibility).

One thing we talked about was whether we should have actually _not_ separated the `header` and the `bottom-bar` into separate components with their own slots, and instead combine them now to make this main component.  So `d2l-navigation` would have a `left`, `right` and `bottom` slot, with a band and shadow, and could be used for _both_ the main nav and the immersive nav, by just choosing whether or not to include anything in the `bottom` slot.

I think one thing that will help is talking to Jeff when he gets back from vacation (I think tomorrow) to get an idea of how rigid he wants the nav to be.  @dlockhart, not sure if you have some thoughts on this as well?

dlockhart: There's definitely a lot of different ways we can go! For now, since we're not totally sure about all the permutations we'll want (at least until Jeff weighs in) and we want to support the existing nav, immersive, plus potentially some edge cases like Parent Portal (and maybe Capture one day?), I like starting flexible and then tightening things up later if it becomes necessary.

We can also have the flexible components AND more rigid ones for the most common cases when and if they emerge.

Also, if we went with the 3-slot approach, I'm not sure how easy it is to detect the fact that they haven't used the `bottom` slot and therefore hide the separator?

test/navigation.html
jeroach: Good call, I'll see if it's possible with ids

d2l-navigation.html
dbatiste: My current thinking is - create the building blocks of nav that could be used different combinations/layouts, and then more rigid components that encapsulate specific layout and responsiveness.  As you said though, we can expose different degrees of freedom.

If we do two different components (one for the 2-slot case, and one for the 3-slot case), then the implementation ends up being simpler.  We _might_ be able to combine them in this case, but if the layout were different, it might not be so simple.  So, keeping different layouts in different components makes a lot of sense to me in this regard.  What does it mean for the consumer though? i.e. is there a case to be made for the consumer to use just one layout component?

That being said, the 3-slot approach is definitely possible with a couple different solutions. 
 Besides the separator, there may also be unnecessary whitespace if there is not content in the bottom slot.  

The approach that I took for `d2l-card` footer slot was to use the resize observer [like this](https://github.com/BrightspaceUI/card/blob/master/d2l-card.html#L262) to check the height of the distributed nodes. It assumes a single distributed node.  It has the advantage (over dom-change/slot-change) of supporting a hidden distributed node - an approach that seems to be used often with the HM components.

Another solution is to wrap the separator and whitespace in a "footer container" component that the consumer must use.  In effect, the consumer is responsible for composing.  It's seemingly pure CSS, but not really because the extra component has JS boilerplate.  This approach provides the advantage of not forcing the whitespace and separator on the consumer.  I guess it depends on how much composition we want the consumer to do, and how much consistency we are hoping to enforce.

jeroach: Thanks for all the discussion everyone! I'm leaning towards keeping the component as is for this PR (as it gives us flexibility, as per previous offline discussion with Dave L and Dave H - seriously you guys need to go by last name or something :P ).

If Jeff is inclined to have more consistent styling enforced (at the cost of flexibility) then I'd be in favour of adding a `d2l-navigation-main` and/or `d2l-immersive-nav` (as suggested by Stacey) in a follow up PR.

I've also removed the unnecessary styling from `d2l-navigation`

test/navigation.html
svanherk: You should be able to grab them by class name, with something like this: https://github.com/BrightspaceUI/inputs/blob/master/test/d2l-input-search.html#L200

d2l-navigation.html
```diff
@@ -0,0 +1,46 @@
...
+<!--
+`d2l-navigation`
+Polymer-based web component for the all encompasing navigation wrapper component
+
+@demo demo/navigation.html
+-->
+
+<dom-module id="d2l-navigation">
+	<template strip-whitespace>
+		<style include="d2l-navigation-shared-styles">
```
svanherk: I think you either need to put this import back at the top, or remove the include here (probably this one).

jeroach: Whoops, yep - thanks :)

test/navigation.html
jeroach: Added :)

svanherk:
Oh actually, did find an issue!  You need to add the new file to `all-imports.html`!

PR #23, opened 2018-11-01
Renaming components as per design discussion

jeroach:
@dbatiste @dlockhart FYI as per our design discussion yesterday 😄


all-imports.html
```diff
@@ -1,7 +1,7 @@
 <link rel="import" href="./d2l-navigation.html">
 <link rel="import" href="./d2l-navigation-band.html">
-<link rel="import" href="./d2l-navigation-header.html">
-<link rel="import" href="./d2l-navigation-bottom-bar.html">
+<link rel="import" href="./d2l-navigation-main-header.html">
+<link rel="import" href="./d2l-navigation-main-footer.html">
```
mtjandra: After merging ensure that the release is a **major** release bump, this will be a breaking change because some places already reference `d2l-navigation-header` and `d2l-navigation-bottom-bar`

PR #26, opened 2018-11-23
add branding colour variables to highlights

jwong-d2l:


jwong-d2l:
I talked with Jeff G, and updated the PR to allow consumers of the navigation to set the color, if they want to.

d2l-navigation-highlight-styles.html
```diff
@@ -22,7 +22,7 @@
 					text-decoration: none;
 				}
 				--d2l-navigation-highlight-base-hover-focus: {
-					color: var(--d2l-color-celestine);
+					color: var(--d2l-navigation-primary-color, var(--d2l-color-celestine));
 					--d2l-icon-fill-color: var(--d2l-color-celestine);
```
svanherk: Do you also want the icon colour to match?

svanherk:
@jeroach We should add this to the README, while you're there:
```
--d2l-branding-primary-color: used to customize the colour of the top navigation band.
--d2l-navigation-primary-color: used to customize the hover colour of the highlight links and buttons
```

d2l-navigation-highlight-styles.html
svanherk: If there's not a use case right now, we could leave it as blue and expand later?

jwong-d2l: I think this does make it match the icon.. at least it did in testing.  We'd like it for our immersive ASV.

svanherk: Oh really?  Are you setting the icon colour with `--d2l-icon-fill-color` yourself?

jwong-d2l: maybe that's a different icon?  I'll check now

jwong-d2l: good call, I checked it and it wasn't working anymore.  strange.

jwong-d2l:
revisit after polymer 3!

PR #29, opened 2018-11-28
Jroach/readme updates

jeroach:
Okay, this is finally at a stage where I'm somewhat happy with it.

For review - it's much easier to just view the rendered readme:
https://github.com/BrightspaceUI/navigation/blob/jroach/readme_updates/README.md

I'm still not sure about the screenshots - part of me thinks that I should put an outline on them (dotted black/grey line or something). It's tricky since the are mostly white at the edges, and the page background is white so it's kind of hard to see where the component ends.

Also, once we're good with this, and all work is wrapped on buttons, then we can update the `Secondary Components` section with the `buttons` and `links`.

Thank you both!


README.md
```diff
@@ -3,7 +3,11 @@
 [![Bower version][bower-image]][bower-url]
 [![Build status][ci-image]][ci-url]
 
-Parent repo for all navigation based web components
+A series of [Polymer](https://www.polymer-project.org/1.0/)-based web components for top level navigation use on D2L pages.
+
+![screenshot of sample usage](/screenshots/navigation.png?raw=true)
+
+For further information on this and other Brightspace UI components, see the docs at [ui.developers.brightspace.com](http://ui.developers.brightspace.com/).
```
svanherk: We don't actually have any info about it here, and the `design.d2l` link isn't up yet (Jeff's getting close though)

README.md
```diff
@@ -23,7 +31,7 @@ Include the [webcomponents.js](http://webcomponents.org/polyfills/) "lite" polyf
 </head>
```
svanherk: They'll technically need to import `d2l-navigation-main-header` and `d2l-navigation-main-footer` here as well, not just the main component

README.md
```diff
@@ -92,6 +114,110 @@ Solid colour band that runs along the top of the navigational header.
 <d2l-navigation-band></d2l-navigation-band>
 ```
 
+***Relevant CSS class names:***
+* `--d2l-branding-primary-color`: Used to customize the colour of the top navigation band.
+* `--d2l-navigation-primary-color`: Used to customize the hover colour of the highlight links and buttons
```
svanherk: This one is technically related to highlight buttons, only the first one relates to the band.  Should we just make a `Buttons and Links` section for now and move that in there?

jeroach: Right - good catch - will update. Thanks!

PR #34, opened 2018-12-03
Add aria-describedby to the button

svanherk:
This fixes the NVDA screenreader problem we were seeing where the notification text was ignored.  It unfortunately only works in Shady DOM, because in Shadow DOM the button cannot see the `d2l-navigation-button`'s slot contents.

### DIFF
--- a/d2l-navigation-button.html
+++ b/d2l-navigation-button.html
@@ -24,14 +24,14 @@ Polymer-based web component for buttons used in the navigational header.
 			{
 				@apply --d2l-navigation-highlight-base-hover-focus;
 			}
-			/* 	
-				::slotted styles for Polymer 1.0; styling all slotted children needs 
+			/*
+				::slotted styles for Polymer 1.0; styling all slotted children needs
 				to be applied explicitely.
-				This cannot be combined with the style block above, as this is not 
-				valid in 2.0 and as such the entire block gets ignored. 
+				This cannot be combined with the style block above, as this is not
+				valid in 2.0 and as such the entire block gets ignored.
 			*/
 			:host(:not([disabled])) button:hover ::slotted(*) *,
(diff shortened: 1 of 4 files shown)


svanherk:
Going to play around with moving the span, to see if I can get this working in Shadow DOM as well

d2l-navigation-button.html
```diff
@@ -65,7 +74,8 @@
 			type$=[[type]]>
 			<span class="d2l-navigation-button-top-border"></span>
 			<slot></slot>
-		</button>
+		</button>
+		<span id="aria-description" class="d2l-offscreen-description">[[ariaDescribedbyText]]</span>
```
jeroach: This appears to be no longer the child element of a `span` with `class` = `d2l-navigation-button-notification-indicator`

Could some style be missing? (I'm going to check BSI)

svanherk: I don't believe so, I think all those styles in that link were needed for the notification dot, not the span itself, since we just move it offscreen

svanherk:
Still working on this, as now I've broken it in shady DOM 😒 .  Going to add a unique id and fix up the tests

d2l-navigation-button.html
CarolBusman: How are we planning on hiding this now when the button is clicked on (i.e. there are no longer any "new" notifications)?

svanherk: Good point, I was assuming the LE would handle that but you're right, it doesn't.  Will add

svanherk:
Ok, this fixes the issue in both Shady DOM and Shadow DOM!  I've only been testing with the demo page, if I have some time tomorrow I'll setup local BSI and confirm this fixes in the LE

svanherk:
Finally had a free second to test this today, works in the LE!

PR #35, opened 2018-12-04
Polymer 3 Conversion

jeroach:


demo/d2l-navigation-button.html
```diff
@@ -4,96 +4,135 @@
...
+function CycleLanguage() {
+	if (currLang >= langCycle.length - 1) {
+		currLang = 0;
+	} else {
+		currLang++;
+	}
+	document.getElementsByTagName('html')[0].setAttribute('lang', langCycle[currLang]);
+}
+function AssignClickEvent(className, handler) {
+	var elems = document.getElementsByClassName(className);
```
svanherk: Only potential issue I see is around this function.  Previously we've needed to wrap calls to elements in an `window.RequestAnimationFrame`, but maybe the `window.onLoad` happens late enough?  Does this stuff still work in the demo?

svanherk:
@dlockhart So the plan was to get Josh to try out the conversion and to leave this branch up in case you or others want to test with the Polymer 3 nav, but we were hoping to get the immersive nav components in before doing the actual merged conversion, since we are adding the immersive nav to legacy pages now.  Let us know if this is an issue!

dlockhart:
Sounds like a plan!

jeroach:
Closing PR for now, but leaving branch.

Will update once `d2l-immersive-nav` is ready as a hybrid component

PR #36, opened 2018-12-06
d2l-navigation-immersive

jeroach:
Ready for initial review. Below is a screenshot of what the component looks like in a few states:

d2l-navigation-immersive_demo

Design doc: http://design.d2l/components/immersive-navbar/

Please note that a lot of the design is out of scope for the first version (e.g. impersonation image) and the middle and right slots are intended to be filled out by the user of the component. So the title, separator, etc. are not part of the base component.

Still to do:
- unit tests
- readme updates

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,14 +16,16 @@
     "wct.conf.json"
   ],
   "dependencies": {
+    "d2l-button": "^4.8.0",
     "d2l-colors": "^3.1.2",
-    "polymer": "Polymer/polymer#1 - 2",
+    "d2l-fastdom-import": "https://github.com/Brightspace/fastdom-import.git#v1.0.2",
     "d2l-icons": "^5.0.2",
+    "d2l-link": "^4.0.2",
+    "d2l-localize-behavior": "^1.1.3",
     "d2l-offscreen": "^3.0.6",
     "d2l-polymer-behaviors": "^1.7.3",
-    "d2l-button": "^4.8.0",
-    "d2l-link": "^4.0.2",
-    "d2l-localize-behavior": "^1.1.3"
+    "d2l-resize-observer-polyfill-import": "https://github.com/Brightspace/resize-observer-polyfill-import.git#^1.5.1",
+    "polymer": "Polymer/polymer#1 - 2"
   },
   "devDependencies": {
     "d2l-typography": "BrightspaceUI/typography#^6.0.0",
--- a/bower.json
+++ b/bower.json
@@ -25,11 +25,10 @@
     "d2l-offscreen": "^3.0.6",
     "d2l-polymer-behaviors": "^1.7.3",
     "d2l-resize-observer-polyfill-import": "https://github.com/Brightspace/resize-observer-polyfill-import.git#^1.5.1",
-    "polymer": "Polymer/polymer#1 - 2",
-    "d2l-typography": "^6.1.3"
+    "d2l-typography": "^6.1.3",
+    "polymer": "Polymer/polymer#1 - 2"
   },
   "devDependencies": {
-    "d2l-typography": "BrightspaceUI/typography#^6.0.0",
     "iron-component-page": "PolymerElements/iron-component-page#^2.0.0",
     "iron-demo-helpers": "PolymerElements/iron-demo-helpers#^2.0.0",
     "iron-test-helpers": "PolymerElements/iron-test-helpers#^2.0.0",
--- a/bower.json
+++ b/bower.json
@@ -25,7 +25,8 @@
     "d2l-offscreen": "^3.0.6",
     "d2l-polymer-behaviors": "^1.7.3",
     "d2l-resize-observer-polyfill-import": "https://github.com/Brightspace/resize-observer-polyfill-import.git#^1.5.1",
-    "polymer": "Polymer/polymer#1 - 2"
+    "polymer": "Polymer/polymer#1 - 2",
+    "d2l-typography": "^6.1.3"
   },
   "devDependencies": {
     "d2l-typography": "BrightspaceUI/typography#^6.0.0",
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -1,5 +1,7 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../d2l-colors/d2l-colors.html">
+<link rel="import" href="../d2l-fastdom-import/fastdom.html">
+<link rel="import" href="../d2l-resize-observer-polyfill-import/resize-observer.html">
 <link rel="import" href="d2l-navigation.html">
 <link rel="import" href="d2l-navigation-link-back.html">
 
@@ -46,18 +48,22 @@ Polymer-based web component for the immersive navigation component
 			}
 
 			.d2l-navigation-immersive-middle {
-				flex: 0 1 auto;
-				font-size: 19px;
-				line-height: 28px;
-				width: 100%
-			}
-
-			.d2l-navigation-immersive-separator {
 				border-left-color: var(--d2l-color-sylvite);
 				border-left-style: solid;
 				border-left-width: 1px;
-				flex: 0 0 auto;
+				border-right-color: var(--d2l-color-sylvite);
+				border-right-style: solid;
+				border-right-width: 1px;
+				flex: 0 1 auto;
+				font-size: 19px;
+				line-height: 28px;
 				margin: 0px 24px 0 24px;
+				padding: 0px 24px 0 24px;
+				width: 100%
+			}
+
+			.d2l-navigation-immersive-middle-hidden {
+				display: none;
 			}
 
 			@media (max-width: 767px) {
@@ -71,8 +77,9 @@ Polymer-based web component for the immersive navigation component
 					height: 60px;
 					margin: 0px 18px 0 18px;
 				}
-				.d2l-navigation-immersive-separator {
+				.d2l-navigation-immersive-middle {
 					margin: 0px 18px 0 18px;
+					padding: 0px 18px 0 18px;
 				}
 			}
 
@@ -81,9 +88,6 @@ Polymer-based web component for the immersive navigation component
 					margin-right: 36px;
 				}
 				.d2l-navigation-immersive-middle {
-					display: none;
-				}
-				.d2l-navigation-immersive-separator {
 					display: none;
 				}
 			}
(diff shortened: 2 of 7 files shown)


jeroach:
(adding @svanherk as an FYI for when you get back)

bower.json
```diff
@@ -16,14 +16,16 @@
     "wct.conf.json"
   ],
   "dependencies": {
+    "d2l-button": "^4.8.0",
```
jeroach: Part of this was simply organizing the dependencies alphabetically

d2l-navigation-immersive.html
```diff
@@ -0,0 +1,175 @@
...
+				</div>
+			</div>
+		</d2l-navigation>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-navigation-immersive',
+
+			properties: {
+				backbuttonhref: {
```
dbatiste: What API do we want on the component?  If we want it to be `<d2l-navigation-immersive back-button-href="">` then we'd want this to be `backButtonHref`.  Usually I think we break up multi-word properties.

dbatiste:
I'm curious about the desired spacing between the components on the RHS.  Have we thought about the spacing... i.e. are we worried about inconsistency, and if so, how we want to mitigate that?

d2l-navigation-immersive.html
dbatiste: A second look at this... seems like we're mixing semantics here.  Internally it's a `d2l-navigation-link-back`, which indeed sounds like it should have an `href` (because it's a link), yet the property we expose on this component (`backbuttonhref`) makes it sound like a button.

dbatiste: Also, is the idea that the left slot contains default element `<d2l-navigation-link-back>`, but the consumer could override it?  If they do provide some other content for the left slot, it seems like the `backbuttonhref` property serves no purpose on the API... is that correct?

d2l-navigation-immersive.html
```diff
@@ -0,0 +1,175 @@
...
+				var middle = Polymer.dom(this.root).querySelector('.d2l-navigation-immersive-middle');
+				if (this._middleObserver) {
+					this._middleObserver.unobserve(middle);
+				}
+			},
+
+			_onMiddleResize: function() {
+				var middleContainer = Polymer.dom(this.root).querySelector('.d2l-navigation-immersive-middle');
+
+				if (middleContainer.clientHeight < 1) {
```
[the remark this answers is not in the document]
jeroach: Hmmm, I used the format I did since I couldn't figure out how to get the slot info from the `entries` object on `d2l-navigation-immersive-middle`

I guess there are 2 ways to go about it:
* Use `fastdom.measure` with my current method
* Change my method to use `entries`

Any thoughts on which is preferred? And any idea how to access elements within a slot off of the `entries` object?

d2l-navigation-immersive.html
```diff
@@ -0,0 +1,175 @@
...
+
+			.d2l-navigation-immersive-left ::slotted(*),
+			.d2l-navigation-immersive-middle ::slotted(*),
+			.d2l-navigation-immersive-right ::slotted(*) {
+				height: 100%;
+			}
+
+			.d2l-navigation-immersive-left {
+				color: var(--d2l-color-tungsten);
+				font-size: 14px;
```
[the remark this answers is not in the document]
jeroach: The designs on `design.d2l` are in `px` - is the onus on developers to convert? Or should the designs be specified in `rem`

@geurts Any thoughts?

d2l-navigation-immersive.html
jeroach: If memory serves (from my offline chat with @geurts ) the intention is that the user _can_ override the left side, but if they do not, then they must provide:
* Text for the back button in the form `Back to <page>`
* `href` for the page URL

Also, I've had issues referencing component attributes when camel case is used. Not sure if I was just doing it wrong or if something else is going on.

I had not considered the API side of this. Do we have any general documentation on the API side of a web component?

jeroach:
@geurts Any thoughts on spacing between components on the RHS? If I recall, we were discussing that the user implementing the `d2l-navigation-immersive` component would be responsible for the inner contents of the right slot and ensuring that it was laid out correctly.

d2l-navigation-immersive.html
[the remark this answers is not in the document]
dbatiste: Ideally we could just use the `entries`, since the browser will have already calculated the dimension for the element being observed (the wrapper around the slot).  Using the `entries` alone offers best performance.  Hopefully we don't need to poke at the contents of the slot, but that's certainly possible if necessary - for example by assignedNodes.

d2l-navigation-immersive.html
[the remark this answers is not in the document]
geurts: I specified fonts using the styles from [Typography](http://design.d2l/style/typography/). In this case, it was "Small Body Text".  

Meanwhile, I specified the spacing using pixels. We generally do it this way, but I do think this is kind of lazy - ideally designers would determine which parts of our layouts should scale with the font size and use REMs for those. In practice, this will be a real challenge to introduce in the design process, but I will bring it up for discussion!

For this case in particular:  it seems the main navigation bar only scales the height of the Link area, while the space between links (and everything in the upper navbar) remains constant. Let's do the same here and have the height specified in REMs so that it's 3.3 REMs instead of 66px (and 3 REMs once we hit the breakpoint).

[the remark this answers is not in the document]
dbatiste: I think we have the `px/rem` listed as an agenda item for the dev/design working group, so we can chat about how to best tackle this.  :)

d2l-navigation-immersive.html
amauryperez: Is there a reason why we can't just definer the properties as `back-link-text` and `back-link-link`?

amauryperez: When I was working on the basic component or now in the Lms wrapper I will have to reference them like that because if I try to use `backLinkHref` or `backLinkText` it just does not find them anyways

amauryperez: Or was is wrong with just `href` and `text`?

svanherk: It's a Polymer convention, it maps the attributes to dashes to the camel-case properties.  When using the component in the LMS, you'll use `back-link-text` and `back-link-link`.

dbatiste: As per the Polymer specs: the property `fooBar` will become the attribute `foo-bar`.  Polymer has code that handles this.  As a practice, we want multi-word property to be camel case on our web components.  Hyphens in **property** names are invalid.  When you render the MVC wrapper, you are rendering attributes. ex.
```
<d2l-some-wrapper foo-bar="stuff">
```

I suggest the simplest names that are needed for the component.  In this case, this is a composite component, more than just a button or a link, so I do not think `href` and `text` sufficiently describe their purpose.

amauryperez: Sounds good

dbatiste: If the camel case names are not working, then something doesn't seem right.

d2l-navigation-immersive.html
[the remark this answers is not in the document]
jeroach: Updated to use just `entries`

d2l-navigation-immersive.html
```diff
@@ -0,0 +1,194 @@
...
+					fastdom.mutate(function() {
+						// nothing in middle
+						middleContainer.classList.add('d2l-navigation-immersive-middle-hidden');
+					});
+					return;
+				}
+
+				var elementInSlot = slotContent[0];
+
+				if (!elementInSlot || elementInSlot.clientHeight < 1) {
```
dbatiste: This is referencing the standard `clientHeight` on a DOM element (from assigned nodes of `entry.target.childNodes[0]`) and will result in a standard measurement.  To get the full benefit of the resize observer, we need to use the dimensions (if possible) of the entries object: `entries[i].contentRect.width`, since the dimensions provides in the entries are already calculated.

d2l-navigation-immersive.html
```diff
@@ -0,0 +1,188 @@
...
+-->
+
+<dom-module id="d2l-navigation-immersive">
+	<template strip-whitespace>
+		<style>
+			d2l-navigation {
+				border-bottom: 1px solid var(--d2l-color-mica);
+			}
+			.d2l-navigation-immersive-margin {
+				margin: 0px 30px;
```
amauryperez: can we make this `margin: 0px 20px;` that will allow for this
image

otherwise we get

image

svanherk: That goes against the design (http://design.d2l/components/immersive-navbar/).  I'd say we're better off changing the iterator's margins

svanherk: But the margin changes at different screen widths anyways, so it'll either always be different in some cases, or we need to implement the same growing/shrinking logic in the assessment page.

amauryperez: 👍

d2l-navigation-immersive.html
jeroach: hmmm, thanks for letting me know. I think we're hosed due to the slot, but I'll give it another crack.

dbatiste: card had a slot too.  might need to check the width too.

bower.json
```diff
@@ -16,14 +16,17 @@
...
+    "d2l-link": "^4.0.2",
+    "d2l-localize-behavior": "^1.1.3",
     "d2l-offscreen": "^3.0.6",
     "d2l-polymer-behaviors": "^1.7.3",
-    "d2l-button": "^4.8.0",
-    "d2l-link": "^4.0.2",
-    "d2l-localize-behavior": "^1.1.3"
+    "d2l-resize-observer-polyfill-import": "https://github.com/Brightspace/resize-observer-polyfill-import.git#^1.5.1",
+    "polymer": "Polymer/polymer#1 - 2",
+    "d2l-typography": "^6.1.3"
```
dbatiste: nit: we usually try to keep at least the d2l deps ordered.

bower.json
```diff
@@ -16,14 +16,17 @@
...
   },
   "devDependencies": {
     "d2l-typography": "BrightspaceUI/typography#^6.0.0",
```
dbatiste: Looks like d2l-typography is listed as both a dev and prod dependency.

d2l-navigation-immersive.html
```diff
@@ -0,0 +1,197 @@
...
+			_onMiddleResize: function() {
+				var middleContainer = Polymer.dom(this.root).querySelector('.d2l-navigation-immersive-middle');
+
+				fastdom.measure(function() {
+					if (middleContainer.clientHeight < 1) {
+						return;
+					}
+				});
+
+				var slotElement = Polymer.dom(this.root).querySelector('.d2l-navigation-immersive-middle > slot');
```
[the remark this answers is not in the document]
jeroach: I'll work on this.

FYI @dbatiste - given this work is blocking others on our team, the plan discussed at our standup this morning was to merge this PR as is (if we're okay with the component as it is).

I'll then continue to work on the unit tests and this refactoring in a follow up PR.

Please let us know if you have any concerns with this. Thanks!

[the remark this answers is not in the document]
dbatiste: Sounds great.  :)

d2l-navigation-link-back.html
```diff
@@ -24,7 +24,8 @@
 				color: inherit;
 			}
 			span {
-				padding: 0 5px;
+				padding: 0;
```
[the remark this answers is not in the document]
svanherk: Yup you're right, not needed.  Good catch!

demo/d2l-navigation-button.html
```diff
@@ -26,6 +26,62 @@
 				margin: 0px 5px;
```
svanherk: Forgot to mention before, but all the changes in this file were just to get the button demo working in Edge, in a way that will be handled properly by the Polymer 3 converter.

jeroach:
FYI follow up work taking place here: https://github.com/BrightspaceUI/navigation/tree/jroach/immersive-follow-up

PR #41, opened 2019-01-02
Make d2l-navigation-immersive sticky

svanherk:
This is a result of an issue we were seeing when trying to use the immersive nav in the Assignment Evaluations page, which has a fixed header.  Even with a `z-index` of `-100`, the gradient was overlappign the content below because the whole header is fixed. This is going to be an issue with everyone that wants the immersive nav to be sticky.  Since this is actually in the spec (http://design.d2l/components/immersive-navbar/) that the nav always be sticky, I think we should handle it here.  This PR:
- Adds `position: fixed` (not `sticky`, since that is not supported in IE11) to the immersive nav, sticking it to the top of the page
- Adjusts the demo page to be able to show and test this properly
- Exposes a new css variable `--d2l-navigation-shadow-gradient-display` that allows for the gradient to be hidden in `d2l-navigation` (needed for the grades page in Shadow DOM, and for this solution)

### DIFF
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -6,6 +6,7 @@
 <link rel="import" href="../d2l-typography/d2l-typography-shared-styles.html">
 <link rel="import" href="d2l-navigation.html">
 <link rel="import" href="d2l-navigation-link-back.html">
+<link rel="import" href="d2l-navigation-shared-styles.html">
 
 <!--
 `d2l-navigation-immersive`
@@ -16,8 +17,17 @@ Polymer-based web component for the immersive navigation component
 
 <dom-module id="d2l-navigation-immersive">
 	<template strip-whitespace>
-		<style>
+		<style include="d2l-navigation-shared-styles">
+			.d2l-navigiation-immersive-fixed {
+				background-color: white;
+				left: 0px;
+				position: fixed;
+				right: 0px;
+				top: 0px;
+				z-index: 1;
+			}
 			d2l-navigation {
+				--d2l-navigation-shadow-gradient-display: none;
 				border-bottom: 1px solid var(--d2l-color-mica);
 			}
 			.d2l-navigation-immersive-margin {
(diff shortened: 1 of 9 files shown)


d2l-navigation-immersive.html
```diff
@@ -16,8 +17,17 @@
...
+			.d2l-navigiation-immersive-fixed {
+				background-color: white;
+				left: 0px;
+				position: fixed;
+				right: 0px;
+				top: 0px;
+				z-index: 1;
+			}
 			d2l-navigation {
+				--d2l-navigation-shadow-gradient-display: none;
```
svanherk: Will update the README to talk about this

d2l-navigation-immersive.html
```diff
@@ -72,6 +82,13 @@
 				--d2l-navigation-link-back-left-padding: 12px;
 			}
 
+			.d2l-navigation-shadow-gradient {
```
svanherk: By removing the gradient from the overall `fixed` div and placing `fixed` on it separately, the gradient is now properly appearing behind the contents below the header.

demo/navigation-immersive-demos/navigation-immersive-all-slots.html
```diff
@@ -0,0 +1,67 @@
...
+		<link rel="import" href="../../../iron-demo-helpers/demo-pages-shared-styles.html">
+		<link rel="import" href="../../../iron-demo-helpers/demo-snippet.html">
+		<link rel="import" href="../../../d2l-typography/d2l-typography.html">
+		<link rel="import" href="../../../d2l-typography/d2l-typography-shared-styles.html">
+		<link rel="import" href="../../d2l-navigation-button.html">
+		<link rel="import" href="../../d2l-navigation-button-close.html">
+		<link rel="import" href="../../d2l-navigation-immersive.html">
+		<custom-style>
+			<style is="custom-style" include="demo-pages-shared-styles">
+				body {
```
svanherk: Set one of the three demos to have a pink background, to make it easier to see the gradient

d2l-navigation-immersive.html
```diff
@@ -16,8 +17,17 @@
 
 <dom-module id="d2l-navigation-immersive">
 	<template strip-whitespace>
-		<style>
+		<style include="d2l-navigation-shared-styles">
+			.d2l-navigiation-immersive-fixed {
+				background-color: white;
+				left: 0px;
```
dbatiste: Nit: our practice for 0 values is to omit the units.  I think originally (years ago) it was a microperf optimization.  Now I suspect it's just convention.

PR #42, opened 2019-01-03
Small adjustments and fixes to demos to prep for Polymer 3 Conversion

svanherk:


demo/navigation-immersive-demos/navigation-immersive-no-slots.html
```diff
@@ -6,7 +6,8 @@
 		<title>d2l-navigation-immersive demo</title>
 		<script src="../../../webcomponentsjs/webcomponents-lite.js"></script>
 		<link rel="import" href="../../../iron-demo-helpers/demo-pages-shared-styles.html">
-		<link rel="import" href="../../../iron-demo-helpers/demo-snippet.html">
+		<link rel="import" href="../../../iron-demo-helpers/demo-snippet.html">
+		<link rel="import" href="../../../d2l-typography/d2l-typography.html">
```
jeroach: This adds an import, but it isn't used in this PR. I assume the previous logic was "broken" in that it was missing the import?

dlockhart: Yeah it's using `d2l-typography` down below a few lines. It just happened to be working because typography was included by other things, but not guaranteed to always be the case.

demo/navigation-main-header.html
```diff
@@ -17,7 +17,7 @@
 		</custom-style>
 		<style>
 			html {
-				font-size: 20px;
+				font-size: 18px;
```
dbatiste: Curious what the rationale is for 18px root font-size for the demo as opposed to the 20px?

svanherk: Right now typography in these demos is broken because of that "missing `body`" issue and we never added the hack here.  Polymer 3 fixes this, and once Lato is added the font is a lot larger.  Here I'm jsut to make the demo space wider and the font a bit smaller to make it look a little bit closer to what it was before, so you can see more stages of responsiveness.  Without this change, the font is already wrapping in the first slot and you don't get to see the gutters growing/shrinking automatically.

PR #43, opened 2019-01-04
Add d2l-focusable

dbatiste:


dbatiste:
I was considering tidying up the demo page a bit, but decided to basically leave it as-is.  Ideally I would like to be able to test the focus method for each element, so that would be a bunch of test buttons, and to keep it organized, perhaps it would be good to have a demo-snippet per component, and then one that combines them.  Perhaps a demo page per component would be the best.

svanherk:
Thanks Dave!  I also like the idea of a demo page per component, but I'm thinking that may be easier to do after the Polymer 3 conversion as part of the cleanup?  Otherwise that's a ton of ugly demo files, although they'd probably be pretty small

PR #44, opened 2019-01-04
Add div for spacing of the immersive nav

svanherk:
Right now, consumers need to add a margin to the top of their page when using the immersive nav, so the start of the content is visible. Instead of having them worry about it (and the responsive rules), I'm adding a div here to take up the same amount of space as the header.  This will only work if the immersive nav element is the first visual element on the page, but that _should_ always be the case anyways and I'm good with enforcing that for now.

### DIFF
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -87,11 +87,8 @@ Polymer-based web component for the immersive navigation component
 			}
 
 			.d2l-navigation-immersive-spacing {
-				display: block;
 				height: calc(var(--d2l-navigation-immersive-height-main) + 5px);
-				margin: 0;
 				position: unset;
-				width: 100%;
 			}
 
 			.d2l-navigation-shadow-gradient {
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -18,6 +18,10 @@ Polymer-based web component for the immersive navigation component
 <dom-module id="d2l-navigation-immersive">
 	<template strip-whitespace>
 		<style include="d2l-navigation-shared-styles">
+			:host {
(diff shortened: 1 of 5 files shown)


d2l-navigation-immersive.html
```diff
@@ -82,11 +86,19 @@
...
+				position: unset;
+				width: 100%;
+			}
+
 			.d2l-navigation-shadow-gradient {
 				left: 0;
 				position: fixed;
 				right: 0;
-				top: 3.1rem;
+				top: calc(var(--d2l-navigation-immersive-height-main) + 5px);
```
svanherk: This was technically a bug before, the gradient was starting a little higher than it had to.  The `5px` is to account for the `4px` colour band and `1px` bottom border.

demo/navigation-immersive-demos/navigation-immersive-all-slots.html
```diff
@@ -16,10 +16,12 @@
 			<style is="custom-style" include="demo-pages-shared-styles">
 				body {
 					background-color: pink;
+					padding: 0;
```
svanherk: Overriding some styles the demo pages add automatically, to make the immersive nav start at the top of the page.

d2l-navigation-immersive.html
```diff
@@ -82,11 +86,19 @@
 				--d2l-navigation-link-back-left-padding: 12px;
 			}
 
+			.d2l-navigation-immersive-spacing {
+				display: block;
+				height: calc(var(--d2l-navigation-immersive-height-main) + 5px);
+				margin: 0;
+				position: unset;
+				width: 100%;
```
dlockhart: Technically I think the `display`, `margin` and `width` properties here are all redundant as they're the defaults for `div`s.

PR #46, opened 2019-01-04
Fix package name

svanherk:
I swear I thought we already did this... 😕 

Maybe I'm losing it


dlockhart:
You're not losing it... looks like I only made the change to the [temporary branch though](https://github.com/BrightspaceUI/navigation/commit/32f1dcb197873e25413431c2123df7207aca4e4c).

PR #47, opened 2019-01-08
Remove check for elements with height less than 1

jeroach:
Resolves [DE32394](https://rally1.rallydev.com/#/42960320374d/detail/defect/277379116328?fdp=true)

In light of this comment:
https://github.com/BrightspaceUI/card/blob/master/d2l-card.js#L284

I don't think we need to check for the height less than 1 since we're not making use of `box-sizing: border-box;`

@dbatiste Can you please confirm that my understanding is correct? Thank you!

### DIFF
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -187,7 +187,6 @@ Polymer-based web component for the immersive navigation component
 				} else {
 					slotContent = D2L.Dom.getComposedChildren(middleContainer);
 				}
-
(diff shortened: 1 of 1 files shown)


jeroach:
Looking into failed test

dbatiste:
Checking the height (at least in other implementations) was to ensure the component is visible, since commonly consumers will hide the element based on some criteria, rather than omitting it from the slot.

jeroach:
Closed in favour of: https://github.com/BrightspaceUI/navigation/pull/48

PR #48, opened 2019-01-08
Add Edge specific logic to support hide/show logic for middle slot

jeroach:


d2l-navigation-immersive.html
```diff
@@ -75,7 +75,7 @@
 				flex: 0 1 auto;
 				margin: 0 24px;
 				padding: 0 24px;
-				width: 100%
+				width: 100%;
```
jeroach: Just fixed this typo while I was in the area

jeroach:
I hate Edge.... 😢

jeroach:
Looking into a way to detect Edge for sauce labs test runs

d2l-navigation-immersive.html
```diff
@@ -212,7 +212,17 @@
...
 				fastdom.measure(function() {
-					var slotHeight = elementInSlot.clientHeight;
+					// Internet Explorer 6-11
+					var isIE = /*@cc_on!@*/false || !!document.documentMode;
+					// Edge 20+
+					var isEdge = !isIE && !!window.StyleMedia;
+
+					var slotHeight;
+					if (isEdge) {
+						slotHeight = elementInSlot.height; // Edge doesn't recognize clientHeight
```
dbatiste: How about `elementInSlot.getBoundingClientRect().height`?

jeroach: @dbatiste Doesn't work in Edge and the tests fail

dbatiste: Interesting - we use `getBoundingClientRect` a lot.   :(

jeroach: @CarolBusman `clientHeight` is defined in Edge, but it has value 0. This is when there actually is valid content in the slot. All other browsers recognize that the slot has content with a height > 0. But Edge thinks it's an `UnknownHtmlElement` with height 0 (for some reason).

CarolBusman: So, assuming our options are:
- Valid non-Edge, has clientHeight > 0 height > 0
- Invalid non-Edge, has clientHeight = 0 height = undefined or 0
- Valid Edge, has clientHeight = 0 height > 0
- Invalid Edge, has clientHeight = 0 height = undefined or 0

Would something like 
```
...
var slotHeight = elementInSlot.clientHeight || elementInSlot.height;

if (!elementInSlot || !slotHeight) {
...
```
work?

jeroach: hmmm, this works in practice but the tests still fail locally (on Chrome)

jeroach: Carol and I chatted offline - I missed a line in the potential solution - tests actually work locally (Chrome) but the logic doesn't work in Edge with Polymer 2.

Back to the drawing board.

jeroach: Carol did some more offline digging and realized that `height` and `clientHeight` are always undefined in Edge

AKA the current "solution" isn't one - still broken.

Images I pasted in the comment below further help to outline the issue.

svanherk:
So I think you're trying to measure the height before it's available.  If I add a `setTimeout` to the code inside the second `fastdom.measure`, everything seems to work in the demo on Edge and my Firefox.  That's not a good solution though, maybe we need to `requestAnimationFrame`?

PR #49, opened 2019-01-09
DE32394 - Fix Hiding of Middle Slot

svanherk:
Alternative to the fix we were trying to get working in https://github.com/BrightspaceUI/navigation/pull/48.  Edge is refusing to cooperate with `clientHeight` on a `slot`, so this is a bit of a redesign of the whole system.  We now have a div in the `d2l-navigation-immersive-middle` div that is only used to tell us the height of the slot's contents, and the `ResizeObserver` watches that instead.  We can use the `entries` array of the `ResizeObserver` now (before we couldn't because the `d2l-navigation-immersive-middle` had `height: 100%`, but we are now watching the new `div` with `height: auto`), so the code in `_onMiddleResize` can be simplified.

Still need to check the functionality and the tests on every browser!  But wanted to get the review up and sauce running.

### DIFF
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -1,7 +1,6 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../d2l-colors/d2l-colors.html">
 <link rel="import" href="../d2l-fastdom-import/fastdom.html">
-<link rel="import" href="../d2l-polymer-behaviors/d2l-dom.html">
 <link rel="import" href="../d2l-resize-observer-polyfill-import/resize-observer.html">
 <link rel="import" href="../d2l-typography/d2l-typography-shared-styles.html">
 <link rel="import" href="d2l-navigation.html">
@@ -78,8 +77,12 @@ Polymer-based web component for the immersive navigation component
 				width: 100%
 			}
 
+			.d2l-navigation-immersive-middle-observer {
+				height: auto;
+			}
+
 			.d2l-navigation-immersive-middle-hidden {
-				display: none;
+				visibility: hidden;
 			}
 
 			d2l-navigation-link-back {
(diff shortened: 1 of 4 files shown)


d2l-navigation-immersive.html
```diff
@@ -78,8 +77,12 @@
 				width: 100%
 			}
 
+			.d2l-navigation-immersive-middle-observer {
+				height: auto;
+			}
+
 			.d2l-navigation-immersive-middle-hidden {
-				display: none;
```
svanherk: Using `display: none` means we can't ever undo the adding of the hidden class, because everything under this div will no longer have an observable size and height will always be 0.  `visibility: hidden` works well, I added a test and a demo to make sure we remove the hidden class if height is added back into the slot by unhiding content.

test/navigation-immersive.html
```diff
@@ -22,17 +22,42 @@
...
+				<d2l-navigation-immersive back-link-href="https://www.d2l.com" back-link-text="Back to D2L">
+					<div slot="middle" id="middle-hidden" style="display: none;">
 						<p>Economics 101</p>
 					</div>
 				</d2l-navigation-immersive>
 			</template>
 		</test-fixture>
-		<script>
+		<script>
+			var raf = function(cb) {
```
svanherk: This is to try to delay the asserts until after fastdom is done running 😞.  Grabbed this from tests in `card`, but still had to add a small timeout for IE11 and Edge.

d2l-navigation-immersive.html
```diff
@@ -172,60 +177,37 @@
...
+			_onMiddleResize: function(entries) {
+				if (!entries || entries.length === 0) {
+					return;
 				}
 
-				if (!slotContent || slotContent.length === 0) {
+				var entry = entries[0];
+				var middleContainer = Polymer.dom(this.root).querySelector('.d2l-navigation-immersive-middle');
+
+				if (entry.contentRect.height > 0) {
```
dbatiste: Hopefully the rounding issue that Jeff mentioned for card is not also an issue here: https://github.com/BrightspaceUI/card/blob/master/d2l-card.js#L286

svanherk: I can increase the number to be safe, if their content is less than 1px high it's probably not very useful

jeroach:
fyi I checkout out this branch locally and tested the demo pages in Chrome, Edge, and Firefox (latest).

For both polymer 1 and 2, everything looks to behave correctly 🎉

PR #50, opened 2019-01-09
Dbatiste/fix nav button focus

dbatiste:
* Added `pointer-events: none;` to slotted content to address button in shadow-dom not getting focus when clicked.

* reorganized and simplified selectors.

### DIFF
--- a/d2l-navigation-button.html
+++ b/d2l-navigation-button.html
@@ -15,6 +15,12 @@ Polymer-based web component for buttons used in the navigational header.
 <dom-module id="d2l-navigation-button">
 	<template strip-whitespace>
 		<style is="custom-style" include="d2l-navigation-highlight-styles">
+
+			:host {
+				display: inline-block;
+				height: 100%;
+			}
+
 			:host([disabled]) button {
 				@apply --d2l-navigation-highlight-disabled;
 			}
--- a/d2l-navigation-button.html
+++ b/d2l-navigation-button.html
@@ -18,6 +18,11 @@ Polymer-based web component for buttons used in the navigational header.
 			:host([disabled]) button {
 				@apply --d2l-navigation-highlight-disabled;
 			}
+
+			:host button ::slotted(*) {
+				pointer-events: none;
+			}
+
 			:host(:not([disabled])) button:hover,
 			:host(:not([disabled])) button:focus,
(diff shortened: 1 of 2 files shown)


d2l-navigation-button.html
```diff
@@ -15,46 +15,62 @@
 <dom-module id="d2l-navigation-button">
 	<template strip-whitespace>
 		<style is="custom-style" include="d2l-navigation-highlight-styles">
-			:host([disabled]) button {
```
dbatiste: This was just moved to the end.

d2l-navigation-button.html
```diff
@@ -15,46 +15,62 @@
 <dom-module id="d2l-navigation-button">
 	<template strip-whitespace>
 		<style is="custom-style" include="d2l-navigation-highlight-styles">
-			:host([disabled]) button {
-				@apply --d2l-navigation-highlight-disabled;
+
+			:host {
+				display: inline-block;
```
dbatiste: This is added to be consistent with the rest of the nav button elements.  It does not affect the visual output, since the children are specifying the height anyway.

d2l-navigation-button.html
```diff
@@ -15,46 +15,62 @@
...
+
+			:host {
+				display: inline-block;
+				height: 100%;
+			}
+
+			:host button ::slotted(*) {
+				pointer-events: none;
 			}
-			:host(:not([disabled])) button:hover,
```
dbatiste: I changed these `:host(:not[disabled])` selectors to `:host`.

d2l-navigation-button.html
```diff
@@ -15,46 +15,62 @@
...
 		<style is="custom-style" include="d2l-navigation-highlight-styles">
-			:host([disabled]) button {
-				@apply --d2l-navigation-highlight-disabled;
+
+			:host {
+				display: inline-block;
+				height: 100%;
+			}
+
+			:host button ::slotted(*) {
```
dbatiste: In Chrome, focus was being placed on the `body` when the icon or text is clicked (the ancestor in it's DOM context that is focusable, and if none, then body).  This bypasses that behavior, allowing the button in the shadow-DOM to get the pointer event and focus.

PR #52, opened 2019-01-14
US103275 - Adjustments to the design when the right slot is empty

svanherk:
This is a use case the LE will be using, where the left slot has the back button and the middle slot has a title (until an iterator is eventually added in the future).  Spoke to Jeff G about the design of this and made adjustments accordingly.  The design now looks like this (the right slot takes up no space if it is empty, and the right separator is hidden):
image

~Still to do - Fixing and adding tests~ Done

### DIFF
--- a/demo/navigation-immersive-demos/navigation-immersive-responsive-slot-hiding.html
+++ b/demo/navigation-immersive-demos/navigation-immersive-responsive-slot-hiding.html
@@ -43,6 +43,11 @@
 				display: table-cell;
 				vertical-align: middle;
 			}
+			@media (max-width: 900px) {
+				#right.d2l-responsive[slot="right"] {
+					display: none;
+				}
+			}
 			@media (max-width: 700px) {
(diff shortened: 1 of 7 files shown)


demo/navigation-immersive-demos/navigation-immersive-responsive-slot-hiding.html
```diff
@@ -58,7 +63,7 @@
 						<div id="middle" class="d2l-typography d2l-body-standard d2l-responsive" slot="middle">
 							Economics 101
 						</div>
-						<div slot="right">
+						<div id="right" class="d2l-responsive" slot="right">
```
dbatiste: With shadyCSS, semi-generic class names like this can cause trouble.  Worth considering a change to make qualify this for nav, or perhaps even better immersive nav.

dbatiste: Sorry - long file name - didn't see this was demo.

PR #53, opened 2019-01-15
Polymer 3 Prep

svanherk:
Few issues I found when testing the Polymer 3 conversion
- Fix outline in Edge in Polymer 2 (and 3)
- Fix RTL padding in the navigation-link-back component

### DIFF
--- a/d2l-navigation-button.html
+++ b/d2l-navigation-button.html
@@ -57,6 +57,7 @@ Polymer-based web component for buttons used in the navigational header.
 			button {
 				@apply --d2l-navigation-highlight-base;
 				font-family: inherit;
+				outline: none; /* Needed for Edge, outline setting from css mixin is not being applied in Polymer 2 & 3 */
 			}
 
 			.d2l-navigation-button-top-border {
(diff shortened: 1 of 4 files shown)


d2l-navigation-button.html
```diff
@@ -56,7 +56,8 @@
 
 			button {
 				@apply --d2l-navigation-highlight-base;
-				font-family: inherit;
+				font-family: inherit;
+				outline: none; /* Needed for Edge, outline setting from css mixin is not being applied in Polymer 2 & 3 */
```
svanherk: FYI @dlockhart @margaree 
This is also an issue with `d2l-button`, `d2l-link`, and anything that sets `outline: none;` in a css mixin.  @dbatiste and I investigated and it's bizarre.  We'll need to figure out if we want to apply this fix to all components that need it, or dig around some more and try to find the root cause.

dlockhart: Interesting. I wonder if it's a similar bug to the `font-family: inherit` thing... although that was Polymer 1 as well.

svanherk: It acts like it, except Polymer 1 is fine.  We made sure the other properties in the mixin were being applied properly, it looks to be specific to `outline`.  And we could get the mixin's `outline` to be applied if it was _overriding_ an `outline` style placed directly on element (like we've done here).  So if I changed the `outline` colour to pink in the mixin and used `!important`, then Edge would start listening.  But if I did that without this line here, it ignores it.

PR #55, opened 2019-01-17
Specify lint folders as to not lint node_modules

jeroach:
Solves the issue I was having as described over here:
https://d2l.slack.com/archives/CDJE0AYMB/p1547753191076900

Finally got it 😄


dlockhart:
Does it actually still find lint issues?

jeroach:
@svanherk Any objections to merging this? Thanks!

PR #56, opened 2019-02-05
supporting a slimmer logo for when there's a title

dlockhart:
Replaces [this code in BSI](https://github.com/Brightspace/brightspace-integration/blob/master/sass/navigation/logo.scss#L15), which wasn't working with shadow DOM anyway.

### DIFF
--- a/d2l-navigation-link-image.html
+++ b/d2l-navigation-link-image.html
@@ -17,9 +17,13 @@ Polymer-based web component for the image link used in the navigational header.
 			}
 			img {
 				vertical-align: middle;
-				border: none;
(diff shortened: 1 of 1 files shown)


d2l-navigation-link-image.html
```diff
@@ -41,6 +45,10 @@
 				},
 				href: {
 					type: String
+				},
+				slim: {
```
dbatiste: Nit: might not be needed for the expected use (i.e. setting via property), but I think this should be reflected to the attribute since the CSS will depends on it.

dlockhart: Right, will update.

PR #58, opened 2019-02-07
applying these styles now that the feature flag is always-on

dlockhart:
Moved these styles here [from here in BSI](https://github.com/Brightspace/brightspace-integration/blob/master/sass/navigation/main.scss#L21) now that the branding feature flag is always-ON.

### DIFF
--- a/d2l-navigation-main-footer.html
+++ b/d2l-navigation-main-footer.html
@@ -13,6 +13,8 @@ Polymer-based web component for the smaller, lower section of the navigational e
 	<template strip-whitespace>
 		<style include="d2l-navigation-shared-styles">
 			:host {
+				border-top: 1px solid rgba(124,134,149,0.18);
+				border-bottom: 1px solid rgba(124,134,149,0.18);
 				display: block;
 			}
 		</style>


d2l-navigation-main-footer.html
```diff
@@ -13,6 +13,8 @@
 	<template strip-whitespace>
 		<style include="d2l-navigation-shared-styles">
 			:host {
+				border-top: 1px solid rgba(124,134,149,0.18);
```
svanherk: We already import `d2l-colors` in here, any reason you didn't use `$d2l-color-light-galena`?

dlockhart: Haha, yes, that confused me too. It's because... it doesn't exist... it's [just an artifact of BSI](https://github.com/Brightspace/brightspace-integration/blob/master/sass/navigation/common.scss#L17)

svanherk: Whhhaaat... should we be using a real d2l color?  Strange lol

dlockhart: Haha well... `124, 134, 149` is Galena, but in this case we want some translucency so we're using an alpha of `0.18`. So it's a weird case because it's an `rgba`.

dlockhart: But yeah... that navbar goes on white, so we could have just figured out what colour that actually mapped to and used that -- or ideally used one in our palette that was to that. 🤷‍♂️

svanherk: Yeah, looks like none of the other `d2l-colors` are in rgba, but you're right, I don't think it needed to be translucent. 🤷‍♀️

dbatiste: For what it's worth, looks like this was intentional:  https://d2l.invisionapp.com/share/T8N1669VN9U#/screens/313447943_Style_Details

PR #60, opened 2019-02-12
Align Immersive navigation max-width to content max-width

ctwomey1:
While integrating the immersive nav into Brightspace for Parents, and Portfolio I noticed that the immersive nav wasn't aligning the back button with our page content.  I confirmed with @geurts that it should be aligned with the content.

In this pr I'm adding a css variable to allow the consumer to set the content max width.  Additionally it looks like my vscode normalized the line endings to LF in the files i've touched.

Demo image:
image

### DIFF
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -38,11 +38,11 @@ Polymer-based web component for the immersive navigation component
 			}
 
 			.d2l-navigation-immersive-container {
-				margin: 0 auto;
-				max-width: var(--d2l-navigation-immersive-content-max-width, 100%);
 				display: flex;
 				height: var(--d2l-navigation-immersive-height-main);
 				justify-content: space-between;
+				margin: 0 auto;
+				max-width: var(--d2l-navigation-immersive-content-max-width, 100%);
 				overflow: hidden;
 			}
 
--- a/d2l-navigation-immersive.html
+++ b/d2l-navigation-immersive.html
@@ -34,6 +34,8 @@ Polymer-based web component for the immersive navigation component
 				border-bottom: 1px solid var(--d2l-color-mica);
 			}
 			.d2l-navigation-immersive-margin {
+				display: flex;
+				justify-content: center;
 				margin: 0 30px;
 			}
 
@@ -41,9 +43,10 @@ Polymer-based web component for the immersive navigation component
 				display: flex;
 				height: var(--d2l-navigation-immersive-height-main);
 				justify-content: space-between;
-				margin: 0 auto;
+				margin: 0 -7px;
 				max-width: var(--d2l-navigation-immersive-content-max-width, 100%);
 				overflow: hidden;
+				width: 100%;
 			}
 
 			.d2l-navigation-immersive-left ::slotted(*),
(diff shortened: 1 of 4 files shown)


d2l-navigation-immersive.html
```diff
@@ -38,10 +38,11 @@
 			}
 
 			.d2l-navigation-immersive-container {
+				margin: 0 auto;
```
jstefaniuk-d2l: nit: Classes are alphabetical in these brightspace ui projects

jstefaniuk-d2l:
It's probably worth adding a readme entry for this as well

jstefaniuk-d2l:
LGTM although I'll leave the approval to someone who's familiar with this repo

svanherk:
Hey Chester, I can take a look after lunch!  The Travis failure is a Firefox + Sauce + P1/2 fluke we've been seeing for a while now, I've just been rekicking CI for you

d2l-navigation-immersive.html
```diff
@@ -41,7 +41,8 @@
 				display: flex;
 				height: var(--d2l-navigation-immersive-height-main);
 				justify-content: space-between;
-				margin: 0 -7px;
```
svanherk: So this `-7px` was there to specifically handle the overhang of button highlight on hover

svanherk: It doesn't seem to look bad without it, it was meant to counteract the 7px padding that has to be added so the highlight overhang isn't cut off.  If design is ok with that padding there, I think we're ok with replacing this rule.

ctwomey1: did some rejiggering with @jstefaniuk-d2l and it looks like i can preserve the -7px.

Looking at lower resolutions the -7 margin seems to matter more

svanherk:
I'm not seeing the css variable being applied in the Polymer 1 demo, but that may just be a problem in the demo, not the code

ctwomey1:
Moving the variable into a :root custom style seems to make it work for the demo

demo/navigation-immersive-demos/navigation-immersive-all-slots-max-width.html
```diff
@@ -0,0 +1,76 @@
...
+		<link rel="import" href="../../../iron-demo-helpers/demo-pages-shared-styles.html">
+		<link rel="import" href="../../../iron-demo-helpers/demo-snippet.html">
+		<link rel="import" href="../../../d2l-typography/d2l-typography.html">
+		<link rel="import" href="../../../d2l-typography/d2l-typography-shared-styles.html">
+		<link rel="import" href="../../d2l-navigation-button.html">
+		<link rel="import" href="../../d2l-navigation-button-close.html">
+		<link rel="import" href="../../d2l-navigation-immersive.html">
+		<custom-style>
+			<style is="custom-style" include="demo-pages-shared-styles">
+				:root {
```
dbatiste: I think you probably want `html` instead of `:root`, since `:root` is obsolete.

svanherk: Confirmed that changing this to `html` does still work as expected

PR #61, opened 2019-02-15
Polymer 3 2019 02 15 081000

ctwomey1:
There's more changes present than in my pull request, but it looks like they are mostly dropping the CRLF for LF.


package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "lint": "npm run lint:wc && npm run lint:js",
     "lint:js": "eslint . --ext .js,.html test/**/*.js test/**/*.html demo/**/*.js demo/**/*.html",
-    "lint:wc": "polymer lint -i \"./*\" \"test\" \"demo\"",
+    "lint:wc": "polymer lint",
```
ctwomey1: Should i revert this change back to ` "lint:wc": "polymer lint -i \"./*\" \"test\" \"demo\"",`?

dlockhart: Probably it was something that should have been in the patch but was missed.

svanherk: I'll add this to the patch

PR #68, opened 2019-06-06
Shared iterator

jeroach:
Thanks for all the input in our discussion!

Progress on what we talked about:
- [x] Extract logic for sub component to `d2l-navigation-iterator-item`
- [x] Use `button` instead of `link`
- [x] Use `text` instead of `title` 
- [x] Add `noNext` and `noPrevious`
- [x] Middle part is now `<slot>`
- [x] Height and media queries in the component itself
- [x] Localization support code

To-do:
- [x] Event firing (on button click)
- [x] Translations for "next" and "previous"
- [x] Reverse padding for RTL
- [x] Unit tests

Rally Story (w/ design info): https://rally1.rallydev.com/#/238990806412d/detail/userstory/304835243544


d2l-navigation-iterator.js
```diff
@@ -0,0 +1,127 @@
...
+					</div>
+					<d2l-tooltip for="d2l-navigation-iterator-next" position="bottom">[[_nextTitleToDisplay]]</d2l-tooltip>
+				</d2l-navigation-link>
+			</div>
+		`;
+	}
+
+	static get properties() {
+		return {
+			previousLink: {
```
jeroach: Ideally this (and `nextLink`) would be mandatory - not sure how to enforce that. Didn't see anything here:
https://polymer-library.polymer-project.org/3.0/docs/devguide/properties

demo/d2l-navigation-iterator.html
```diff
@@ -0,0 +1,69 @@
...
+				</custom-style>
+				<custom-style include="d2l-typography">
+					<style is="custom-style" include="d2l-typography"></style>
+				</custom-style>
+				<style>
+					html {
+						font-size: 20px;
+					}
+					.iterator-container {
+						height: 66px;
```
jeroach: Picked a parent height of 66px, since that's the height of the immersive nav

d2l-navigation-iterator.js
```diff
@@ -0,0 +1,127 @@
...
+
+	_getNextTitleToDisplay(nextTitle) {
+		if (nextTitle.length > 0) {
+			return nextTitle;
+		}
+		return 'Next'; // TODO: localize
+	}
+
+	_displayLinkTitles() {
+		if (this.countInformation.length > 0) {
```
AlexBedley: nit: this could just be `return this.countInformation.length <= 0`

Or since it's always negated when it's used you could change it to `_hideLinkTitles` and just keep it `> 0`

demo/d2l-navigation-iterator.html
svanherk: I think that changes at a certain width though, right?

d2l-navigation-iterator.js
```diff
@@ -0,0 +1,127 @@
...
+				type: String,
+				value: '',
+				reflectToAttribute: true
+			},
+			nextLink: {
+				type: String,
+				value: '',
+				reflectToAttribute: true
+			},
+			previousTitle: {
```
AlexBedley: We can get rid of `_previousTitleToDisplay` and `_nextTitleToDisplay`. Just use `previousTitle` and set the default to the localized version of `Previous`.

d2l-navigation-iterator.js
```diff
@@ -0,0 +1,127 @@
...
+				</d2l-navigation-link>
+			</div>
+		`;
+	}
+
+	static get properties() {
+		return {
+			previousLink: {
+				type: String,
+				value: '',
```
AlexBedley: Any particular reason why these defaults exists? (for `previousLink`, `nextLink`, and `countInformation`)

demo/d2l-navigation-iterator.html
jeroach: You are correct - below `615px` it goes down to `60px`
http://design.d2l/components/immersive-navbar/

jeroach: Do we typically add media queries on our demo pages?

Might be easier to have a demo where the iterator is embedded in the immersive nav component - then we get it for free :)

d2l-navigation-iterator.js
```diff
@@ -0,0 +1,127 @@
...
+				}
+				.d2l-navigation-iterator {
+					align-items: center;
+					display: flex;
+					justify-content: space-between;
+					height: 100%;
+					max-width: 20rem;
+				}
+				.d2l-navigation-iterator-previous-link {
+					padding: 0 0 0 1.2rem;
```
AlexBedley: if the default padding is `0` this should just be `padding-left: 1.2rem;` (easier to read the specific padding attributes than parse the general `padding` format)

d2l-navigation-iterator.js
svanherk: So when we were originally spec-ing out an iterator component for CJ to work on, to try to cut down on the repetitiveness here (`nextLink`, `previousLink`, etc) we came up with something like this:
```
<d2l-iterator>
	<d2l-iterator-item slot="prev" href="<something>" title="something" disabled></d2l-iterator-item>
	<d2l-iterator-item slot="next" href="<something>" title="something"></d2l-iterator-item>
</d2l-iterator>
```
Where `d2l-iterator-item` would understand what icon to use depending on what `type` was passed in, and `d2l-iterator` could still control the inside information.  This was trying to be more flexible, might be overkill but cuts the logic in half

The `disabled` is to handle what happens when you're at the first or the last person/item, which I assume you'll need to add here as well (and they wouldn't need an href in that case)

demo/d2l-navigation-iterator.html
svanherk: I'm thinking that should just be in the component itself, so people don't need to worry about that media shift. I guess it depends if we think this will only be in the navbar, someone could add overrides later though

jeroach: Ah I see, yeah I can fix that

d2l-navigation-iterator.js
jeroach: Sounds reasonable, I can refactor to match

d2l-navigation-iterator.js
```diff
@@ -0,0 +1,127 @@
...
+					height: 100%;
+				}
+			</style>
+			<div class="d2l-navigation-iterator">
+				<d2l-navigation-link id="d2l-navigation-iterator-previous" href="[[previousLink]]" class="d2l-navigation-iterator-previous-link d2l-focusable" text="[[_previousTitleToDisplay]]">
+					<div class="d2l-navigation-iterator-link-items">
+						<d2l-icon icon="d2l-tier3:chevron-left-circle"></d2l-icon>
+						<span class="d2l-navigation-iterator-previous-link-text" hidden$="[[!_displayLinkTitles()]]">[[_previousTitleToDisplay]]</span>
+					</div>
+						<d2l-tooltip for="d2l-navigation-iterator-previous" position="bottom">[[_previousTitleToDisplay]]</d2l-tooltip>
```
AlexBedley: `position="bottom"` is the default. Not sure if we require it or we just want to be consistent (if the default ever changed we could get that for free).

demo/d2l-navigation-iterator.html
svanherk: Or you could try to have it just center vertically at whatever height its given, and have it be 100% of its container.  Then people can just override by setting a height if they don't like that

demo/d2l-navigation-iterator.html
```diff
@@ -0,0 +1,69 @@
...
+			document.body.appendChild($_documentContainer.content);
+		</script>
+	</head>
+	<body class="d2l-typography">
+		<div class="vertical-section-container fixedSize">
+			<h3>d2l-navigation-iterator empty demo</h3>
+			<demo-snippet>
+				<template strip-whitespace>
+					<div class="iterator-container">
+						<d2l-navigation-iterator previous-link="https://www.nfl.com" next-link="https://www.d2l.com"></d2l-navigation-iterator>
```
AlexBedley: Not necessary at all, but you could make the iterator demo link to itself with a different fragment (the part after `#`) which would be the current iteration state. This would provide an example of an iterator controller as well.

d2l-navigation-iterator.js
AlexBedley: yeah can't think of any way to enforce it before runtime (we could throw if previousLink isn't a valid url but by then it's probably too late).

d2l-navigation-iterator.js
svanherk: Also we didn't spend any time on naming, so don't feel you need to match that.  I don't mind `d2l-navigation-iterator`

AlexBedley: Hmm I don't quite understand the advantage of `d2l-iterator-item` as a public element. Maybe under the hood it makes sense to reduce code duplication, but it feels like it adds more things developers could get wrong when trying to use an iterator.

d2l-navigation-iterator.js
jeroach: True, will fix

d2l-navigation-iterator.js
jeroach: I would like to achieve some consensus here - perhaps an offline discussion?

Lest I make a whole bunch of changes and then revert them later.

demo/d2l-navigation-iterator.html
jeroach: I had quite a few issues getting everything to line up and center already, so not super keen to mess around with that.

I don't imagine many usages outside of immersive nav. If someone wants to override in future, that's cool.

d2l-navigation-iterator.js
jeroach: will fix

jeroach: Refactor well in progress as per discussion :)

d2l-navigation-iterator.js
jeroach: Not relevant after refactor

d2l-navigation-iterator.js
jeroach: Removed

d2l-navigation-iterator.js
jeroach: Fixed

demo/d2l-navigation-iterator.html
jeroach: Not relevant after refactor

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,141 @@
...
+class D2LNavigationIteratorItem extends PolymerElement {
+	static get template() {
+		return html`
+			<style>
+				:host {
+					display: flex;
+					height: 100%;
+					position: relative;
+				}
+				.d2l-navigation-iterator-item-button-previous {
```
svanherk: Instead of having separate classes and having to set them to a property below, you could just have one class (`.d2l-navigation-iterator-item-button`) and use the attribute in the css:
```
				[type="previous"] .d2l-navigation-iterator-item-button {
					padding-left: 1.5rem;
				}
				[type="next"] .d2l-navigation-iterator-item-button {
					padding-right: 1.5rem;
				}
```

jeroach: Thanks - will do - much cleaner!

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -0,0 +1,68 @@
...
+				.d2l-navigation-iterator {
+					align-items: center;
+					display: flex;
+					justify-content: space-between;
+					height: 100%;
+					max-width: 20rem;
+				}
+				[hidden] {
+					display: none;
+				}
```
dlockhart: I would move this CSS rule into the item's CSS, since that's the element you're trying to hide. i.e. `:host([hidden]) { display: none; }`.

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -0,0 +1,68 @@
...
+			<style>
+				:host {
+					@apply --d2l-body-compact-text;
+				}
+				.d2l-navigation-iterator {
+					align-items: center;
+					display: flex;
+					justify-content: space-between;
+					height: 100%;
+					max-width: 20rem;
```
dlockhart: Is this wrapper element needed, or could it all most to the host?

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -0,0 +1,68 @@
...
+			},
+			noPrevious: {
+				type: Boolean,
+				value: false,
+				reflectToAttribute: true
+			},
+			noNext: {
+				type: Boolean,
+				value: false,
+				reflectToAttribute: true
```
dlockhart: I don't think any of these need to be reflected back to the attribute?

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,133 @@
...
+				<d2l-navigation-button class$="[[_buttonClass]]" text$="[[_displayText]]">
+					<div class="d2l-navigation-iterator-item-items">
+						<template is="dom-if" if="[[_typeIsPrevious]]">
+							<d2l-icon icon$="[[_icon]]"></d2l-icon>
+							<span class$="[[_textClass]]">[[_displayText]]</span>
+						</template>
+						<template is="dom-if" if="[[!_typeIsPrevious]]">
+							<span class$="[[_textClass]]">[[_displayText]]</span>
+							<d2l-icon icon$="[[_icon]]"></d2l-icon>
+						</template>
```
dlockhart: This ends up being a lot of DOM -- could you use computed private attributes for the text and icon instead and remove the `dom-if`s?

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,133 @@
...
+					height: 100%;
+				}
+				.d2l-navigation-iterator-item-text-previous {
+					padding-left: 0.6rem;
+				}
+				.d2l-navigation-iterator-item-text-next {
+					padding-right: 0.6rem;
+				}
+			</style>
+			<div id="d2l-navigation-iterator-item">
```
dlockhart: Again, is this wrapper `<div>` needed?

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
dlockhart: Oh I see, you need the text to show up on one side or the other. In that case -- you can use `flex-direction` to reverse the display order, [similar to what breadcrumbs does here](https://github.com/BrightspaceUI/breadcrumbs/blob/master/d2l-breadcrumb.js#L24).

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,133 @@
...
+
+	_typeChanged(newValue, oldValue) {
+		const validTypes = ['previous', 'next'];
+		if (validTypes.indexOf(newValue) === -1) {
+			this.type = oldValue;
+		}
+	}
+
+	_setupTypeProperties(type, text) {
+		let defaultText;
```
dlockhart: I think all of this can be replaced by using computed properties for the icon and text.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,133 @@
...
+					display: flex;
+					height: 100%;
+					position: relative;
+				}
+				.d2l-navigation-iterator-item-button-previous {
+					padding-left: 1.2rem;
+				}
+				.d2l-navigation-iterator-item-button-next {
+					padding-right: 1.5rem;
+				}
```
dlockhart: Don't forget to reverse right/left padding for RTL.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
jeroach: Nope

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,115 @@
...
+			<style>
+				:host {
+					display: flex;
+					height: 100%;
+					position: relative;
+				}
+				:host([hidden]) {
+					display: none;
+				}
+				:host([type="previous"]) .d2l-navigation-iterator-item-button {
```
dlockhart: I think the `.d2l-navigation-iterator-item-button` CSS class can just go away and be replaced by `d2l-navigation-button`. Shadow DOM is awesome. :)

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,115 @@
...
+				:host([type="next"]) .d2l-navigation-iterator-item-items {
+					flex-direction: row-reverse;
+				}
+				:host([type="previous"]) .d2l-navigation-iterator-item-text {
+					padding-left: 0.6rem;
+				}
+				:host([type="next"]) .d2l-navigation-iterator-item-text {
+					padding-right: 0.6rem;
+				}
+				[hidden] {
```
dlockhart: I'm surprised this is needed -- the text `span` should hide without it since it's a built-in HTML thing as long as you don't change the `display` CSS property, which you don't.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,115 @@
...
+			<d2l-tooltip for="d2l-navigation-iterator-item">[[_displayText]]</d2l-tooltip>
+		`;
+	}
+
+	static get properties() {
+		return {
+			text: {
+				type: String,
+				value: '',
+				reflectToAttribute: true
```
dlockhart: Text shouldn't need to be reflected.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,115 @@
...
+			return 'Previous'; // TODO: localize
+		}
+		return 'Next'; // TODO: localize
+	}
+
+	_computeIcon(type) {
+		if (type === 'previous') {
+			return 'd2l-tier3:chevron-left-circle';
+		}
+		return 'd2l-tier3:chevron-right-circle';
```
dlockhart: Super nit: for simple conditions like this, I like `return (type === 'previous') ? 'd2l-tier3:chevron-left-circle' : 'd2l-tier3:chevron-right-circle';`. 🤷‍♂

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -0,0 +1,64 @@
...
+			<style>
+				:host {
+					@apply --d2l-body-compact-text;
+					align-items: center;
+					display: flex;
+					justify-content: space-between;
+					height: 3.3rem;
+					max-width: 20rem;
+				}
+				@media (max-width: 615px) {
```
dlockhart: This is a new thing I learned: we should be using `rem` units for our responsive rules. That way if folks use larger or smaller fonts, the breakpoints adjust accordingly.

dlockhart: `615px` = `30.75rem`.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
jeroach: It isn't :)

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,118 @@
...
+					display: none;
+				}
+				:host([type="previous"]) d2l-navigation-button {
+					padding-left: 1.2rem;
+				}
+				:host([type="next"]) d2l-navigation-button {
+					padding-right: 1.5rem;
+				}
+				:host([type="previous"]:dir(rtl)) d2l-navigation-button {
+					padding-right: 1.2rem;
```
svanherk: Do you need to set the `padding-left` back to `0` here?  Same with the one below, I'd expect `padding-right` would need to be set back to `0` to override the settings above

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,118 @@
...
+					flex-direction: row-reverse;
+				}
+				:host([type="previous"]) .d2l-navigation-iterator-item-text {
+					padding-left: 0.6rem;
+				}
+				:host([type="next"]) .d2l-navigation-iterator-item-text {
+					padding-right: 0.6rem;
+				}
+				:host([type="previous"]:dir(rtl)) .d2l-navigation-iterator-item-text {
+					padding-right: 0.6rem;
```
svanherk: Same with the RTL styles here

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,118 @@
...
+				}
+				:host([type="next"]:dir(rtl)) .d2l-navigation-iterator-item-text {
+					padding-left: 0.6rem;
+				}
+			</style>
+			<d2l-navigation-button id="d2l-navigation-iterator-item" class="d2l-navigation-iterator-item-button" text="[[_computeText(text, type)]]">
```
svanherk: Is the `id` here used for anything?

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -0,0 +1,84 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import 'd2l-polymer-behaviors/d2l-focusable-behavior.js';
+import 'd2l-icons/d2l-icon.js';
```
svanherk: You can cleanup a bunch of the imports here, now that they've moved to `d2l-navigation-iterator-item`

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -0,0 +1,84 @@
...
+				type: Boolean,
+				value: false
+			}
+		};
+	}
+
+	static get is() { return 'd2l-navigation-iterator'; }
+
+	_dispatchButtonClicked(e) {
+		this.dispatchEvent(new CustomEvent('d2l-navigation-iterator-button-clicked', {
```
svanherk: Do we want to stop the propagation of the original click event, now that's we're sending out this custom one?

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,120 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import {NavigationLocalize} from '../NavigationLocalize.js';
+import 'd2l-polymer-behaviors/d2l-focusable-behavior.js';
```
svanherk: I don't think this is being used for anything?

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,120 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import {NavigationLocalize} from '../NavigationLocalize.js';
+import 'd2l-polymer-behaviors/d2l-focusable-behavior.js';
+import 'd2l-icons/d2l-icon.js';
+import 'd2l-icons/tier3-icons.js';
+import 'd2l-tooltip/d2l-tooltip.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
```
svanherk: I think this line can be removed.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -0,0 +1,120 @@
...
+				:host([type="previous"]) d2l-navigation-button {
+					padding-left: 1.2rem;
+				}
+				:host([type="next"]) d2l-navigation-button {
+					padding-right: 1.5rem;
+				}
+				:host([type="previous"]:dir(rtl)) d2l-navigation-button {
+					padding-right: 1.2rem;
+				}
+				:host([type="next"]:dir(rtl)) d2l-navigation-button {
```
svanherk: Still missing to `0` override here and in the block above

svanherk:
What's here looks goo to me now, although I do thing we should add a section to the README

package.json
```diff
@@ -48,7 +56,10 @@
     "d2l-offscreen": "BrightspaceUI/offscreen#semver:^4",
     "d2l-polymer-behaviors": "Brightspace/d2l-polymer-behaviors-ui#semver:^2",
     "resize-observer-polyfill": "^1.5.0",
+    "d2l-tooltip": "BrightspaceUI/tooltip#semver:^3",
     "d2l-typography": "BrightspaceUI/typography#semver:^7",
-    "@polymer/polymer": "^3.0.0"
+    "@polymer/polymer": "^3.0.0",
+    "merge-stream": "^1.0.1",
```
svanherk: Can these two (`merge-stream` and `require-dir`) move to dev-dependencies?  Since they're only used to locally build the language files?

README.md
```diff
@@ -283,6 +283,55 @@ Then add the `d2l-navigation-main-footer`, and provide elements for the `main` s
...
+
+The iterator button labels can be customized with `previous-text` and `next-text`.
+
+![screenshot of navigation iterator with hidden iterator button text and custom content](/screenshots/d2l-navigation-iterator-custom-buttons.png?raw=true)
+
+```html
+<d2l-navigation-iterator previous-text="Back" next-text="Forward"></d2l-navigation-iterator>
+```
+
+The iterator buttons can be hidden completely with `no-next` or `no-previous`.
```
svanherk: I forget where we landed with this in our discussion... didn't we talk about this `disabling` the buttons, not hiding them?  Or do we need the ability to support both?  This is probably a question for design, would they rather it be there but disabled, or completely removed

PR #69, opened 2019-07-05
US107607 Make Navigation Notification Dot a component

ciss1995:


d2l-navigation-notification.js
```diff
@@ -0,0 +1,58 @@
...
+import { html, PolymerElement } from '@polymer/polymer/polymer-element.js';
+
+/**
+ * @customElement
+ * @polymer
+ */
+class NavigationNotificationIcon extends PolymerElement {
+	static get template() {
+		return html`
+            <style is="custom-style" include="d2l-offscreen-shared-styles">
```
dlockhart: This isn't using offscreen styles so the `include` isn't required here. Actually, neither is the place this was copied from.

d2l-navigation-notification.js
```diff
@@ -0,0 +1,58 @@
...
+                <div class="d2l-navigation-notification-indicator-icon"></div>
+            </span>
+		`;
+	}
+
+	static get is() { return 'd2l-navigation-notification-icon'; }
+
+	static get properties() {
+		return {
+			notification: {
```
dlockhart: I'd recommend using the native `hidden` attribute instead. So consumers would go `<d2l-navigation-notification-icon hidden>` if they didn't want to see it, and would remove that attribute when they wanted it to appear. You'll need to include some CSS to make that happen:

```css
:host([hidden]) { display: none; }
```

d2l-navigation-notification.js
```diff
@@ -0,0 +1,58 @@
...
+                }
+                .d2l-navigation-notification-indicator-icon {
+                    background: var(--d2l-color-primary-accent-indicator);
+                    border: 2px solid white;
+                    height: 10px;
+                    width: 10px;
+                    border-radius: 50%;
+                }
+            </style>
+            <span class="d2l-navigation-notification-indicator">
```
dlockhart: I think you can collapse this `<span>` and the host itself -- like you can accomplish this with 1 less element.

d2l-navigation-notification.js
```diff
@@ -0,0 +1,58 @@
...
+                    display: inline-block;
+                    height: 100%;
+                }
+                :host([notification]) .d2l-navigation-notification-indicator {
+                    display: inline-block;
+                }
+                .d2l-navigation-notification-indicator {
+                    display: none;
+                    height: 100%;
+                    position: absolute;
```
dlockhart: Do we always want all notification dots in the top-right of their parent's? If so, then having the positioning CSS here is OK... but I would expect the parent to be the one to decide where to position this.

ciss1995: Good question! I think that we always want it to be in the top-right of the parent but I am not 100% sure. @svanherk @ctwomey1

d2l-navigation-notification.js
ciss1995: Seems like `d2l-navigation-button-notification-icon` requires it. After removing it, it renders the `notificationText` from `aria-describedby-text="[[_getNotificationText(notification, notificationText)]]"`

d2l-navigation-notification.js
svanherk: I think that's the only pattern we've seen so far with nav notifications, @geurts would know if we ever expect this to change

report.20190705.153023.30128.0.001.json
```diff
@@ -0,0 +1,283 @@
+
```
svanherk: Looks like this was committed by accident

ciss1995: You are right. I will remove it in the next commit.

d2l-navigation-notification.js
dlockhart: That means something somewhere is using offscreen styles but either isn't importing the offscreen shared styles or isn't putting it in its `include`. I think it's actually [https://github.com/BrightspaceUI/navigation/blob/master/d2l-navigation-button.js#L24](d2l-navigation-button) that's at fault.

PR #70, opened 2019-07-08
Classify things

ctwomey1:
I need to do a bit of work on the navigation for Consortiums, however I didn't want to work on code that is just the output of polymer modulizer.

So I took a bit of time to update everything to modern polymer 3.  Most things were straight forward ports, my biggest changes are in d2l-navigation-immersive.js (attached and detached > connected/disconnected callbacks), and in the way styles are shared (exported const instead of include="").

Please let me know what you think.  My hope was that this code would be easier to work with now, and potentially easier to port to lit.


d2l-navigation-button-close.js
```diff
@@ -1,29 +1,65 @@
...
+Polymer-based web component for buttons used in the navigational header.
+This button contains an X icon.
+
+@demo demo/d2l-navigation-button.html d2l-navigation-button-close
+*/
+class D2LNavigationButtonClose extends mixinBehaviors([D2L.PolymerBehaviors.Button.Behavior,
+	D2L.PolymerBehaviors.FocusableBehavior,
+	D2L.PolymerBehaviors.LocalizeBehavior], PolymerElement) {
+	static get is() {
+		return 'd2l-navigation-button-close';
```
dlockhart: Nit: I believe the `is` business is no longer needed (I can't find it in the Polymer 3 docs), so you could just remove these 3 lines and have `d2l-element-name` down where you register it below.

d2l-navigation-button-close.js
```diff
@@ -1,29 +1,65 @@
...
 import 'd2l-icons/d2l-icon.js';
 import 'd2l-icons/tier1-icons.js';
 import 'd2l-button/d2l-button-behavior.js';
 import 'd2l-polymer-behaviors/d2l-focusable-behavior.js';
 import 'd2l-localize-behavior/d2l-localize-behavior.js';
-import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
-const $_documentContainer = document.createElement('template');
 
-$_documentContainer.innerHTML = `<dom-module id="d2l-navigation-button-close">
-	<template strip-whitespace="">
```
dlockhart: Polymer 3 [doesn't appear to have an equivalent](https://polymer-library.polymer-project.org/3.0/docs/devguide/dom-template#strip-whitespace). I'm curious what you're seeing in the browser with these changes -- is the whitespace gone?

d2l-navigation-band.js
```diff
@@ -25,11 +24,8 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-navigation-band">
...
-</dom-module>`;
 
-document.head.appendChild($_documentContainer.content);
-Polymer({
-	is: 'd2l-navigation-band'
-});
+		`;
+	}
+}
+window.customElements.define(D2LNavigationBand.is, D2LNavigationBand);
```
dlockhart: Nit: `window` things are in the global namespace so it's not needed.

d2l-navigation-button.js
```diff
@@ -1,114 +1,107 @@
...
+				type: String
 			}
-
-		</style>
-		<button aria-describedby$="[[_ariaDescriptionId]]" aria-expanded$="[[ariaExpanded]]" aria-haspopup$="[[ariaHaspopup]]" aria-label$="[[text]]" class="d2l-focusable" disabled$="[[disabled]]" autofocus$="[[autofocus]]" form$="[[form]]" formaction$="[[formaction]]" formenctype$="[[formenctype]]" formmethod$="[[formmethod]]" formnovalidate$="[[formnovalidate]]" formtarget$="[[formtarget]]" name$="[[name]]" title$="[[text]]" type$="[[type]]">
+		};
+	}
+	static get template() {
+		return html`
+				${highlightStyles}
```
dlockhart: I'm a little worried that this is going to result in those styles being copied into every single instance of the button. Lit has solved this using constructable stylesheets... but I'm not sure if there's an elegant Polymer 3 solution. We're on the cusp of moving to Lit anyway, so maybe this isn't a big deal.

ctwomey1: You're right that it will be stamped into each shadow root.  But based on this issue https://github.com/Polymer/polymer/issues/4940#issuecomment-371664641 it sounds like browsers do optimize for that.

dlockhart: Sweet, that makes me feel better. :)

d2l-navigation-button-close.js
ctwomey1: According to this https://github.com/Polymer/polymer/issues/4731 it seems like you can just add the attribute to the template.

I didn't specifically notice anything with whitespace, but i wasn't really looking hard either.  Feels like an easy enough thing to add

d2l-navigation-immersive.js
```diff
@@ -1,30 +1,25 @@
...
-
 import 'd2l-colors/d2l-colors.js';
 import 'fastdom/fastdom.js';
 import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
 import './d2l-navigation.js';
 import './d2l-navigation-link-back.js';
-import './d2l-navigation-shared-styles.js';
-import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
+import { navigationSharedSytle } from './d2l-navigation-shared-styles.js';
```
svanherk: Minor: `Style`

PR #71, opened 2019-07-08
US107607 Made notification dot a component

ciss1995:
US107607
https://github.com/BrightspaceUI/navigation/pull/69

### DIFF
--- a/d2l-navigation-button.js
+++ b/d2l-navigation-button.js
@@ -31,7 +31,7 @@ class D2LNavigationButton extends mixinBehaviors([D2L.PolymerBehaviors.Button.Be
 	static get template() {
 		const template = html`
 				${highlightStyles}
-				<style is="custom-style">
+				<style is="custom-style" include="d2l-offscreen-shared-styles">
 
 					:host {
 						display: inline-block;
--- a/d2l-navigation-notification-icon.js
+++ b/d2l-navigation-notification-icon.js
@@ -13,7 +13,7 @@ class NavigationNotificationIcon extends PolymerElement {
                     display: inline-block;
                     height: 100%;
                     position: absolute;
-                    right: calc(-50% - 4px);
+                    right: calc(-100% + 11px);
                     top: calc(-50% + 11px);
                     width: 100%;
                 }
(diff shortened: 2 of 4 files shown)


d2l-navigation-notification-icon.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+class NavigationNotificationIcon extends PolymerElement {
+	static get template() {
+		return html`
+            <style is="custom-style">
+                :host {
+                    display: inline-block;
+                    height: 100%;
+                    position: absolute;
+                    right: calc(-50% - 4px);
```
ciss1995: We would need to change this calculation to allow the component to work with https://github.com/BrightspaceHypermediaComponents/organizations/blob/master/components/d2l-organization-consortium/d2l-organization-consortium-tabs.js. I tried `right: calc(-50% - 11px);`, which seems to work but I am not sure if that reflects the actual padding we want.

edwardg8:
Looks good to me, i'm not sure about positioning of the dot.

d2l-navigation-notification-icon.js
svanherk: So is the plan to make this configurable?  Or can you override the style in d2l-org-tabs as is?

d2l-navigation-button.js
```diff
@@ -31,7 +31,7 @@ class D2LNavigationButton extends mixinBehaviors([D2L.PolymerBehaviors.Button.Be
 	static get template() {
 		const template = html`
 				${highlightStyles}
-				<style is="custom-style">
+				<style is="custom-style" include="d2l-offscreen-shared-styles">
```
svanherk: Nice, looks like this should've always been here

d2l-navigation-notification-icon.js
svanherk: I'm thinking it should be configurable as is, since it's in the `:host` selector

ciss1995:
After discussing it with @svanherk we agreed to change `right: calc(-50% - 4px);` to `right: calc(-100% + 11px);` to allow it to work with `d2l-organization-consortium-tabs`. We will have to do some regression testing on the nav bar.

PR #72, opened 2019-07-08
Slot in navigation band

ctwomey1:
Adding a slot to the navigation band so that we can render the consortium tabs

This allows for this:
navigation-band
navigation

### DIFF
--- a/d2l-navigation-band.js
+++ b/d2l-navigation-band.js
@@ -1,6 +1,8 @@
 import 'd2l-colors/d2l-colors.js';
+import './d2l-navigation-shared-styles';
 
 import {PolymerElement, html} from '@polymer/polymer/polymer-element.js';
+import { navigationSharedStyle } from './d2l-navigation-shared-styles';
 /**
 `d2l-navigation-band`
 Polymer-based web component for a solid colour band that runs along the top of the navigational header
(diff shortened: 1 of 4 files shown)


d2l-navigation.js
```diff
@@ -30,7 +30,7 @@ class D2LNavigation extends PolymerElement {
 				width: 100%;
 			}
 		</style>
-		<d2l-navigation-band></d2l-navigation-band>
+		<d2l-navigation-band><slot name="navigation-band"></slot></d2l-navigation-band>
```
dlockhart: Do we foresee other slots in the future that would be more default than this one? If not, I don't think you need a name on it, it can just be the default slot.

dlockhart: Nevermind, I see what's going on here. :)

d2l-navigation-band.js
```diff
@@ -11,16 +13,19 @@ class D2LNavigationBand extends PolymerElement {
...
 	static get template() {
 		const template = html`
-
+		${navigationSharedStyle}
 		<style>
 			:host {
-				border-top-color: var(--d2l-branding-primary-color, var(--d2l-color-celestine));
-				border-top-style: solid;
-				border-top-width: 4px;
+				background: var(--d2l-branding-primary-color, var(--d2l-color-celestine));
```
dlockhart: Nit: could be more specific `background-color` instead of the complex shorthand `background`.

PR #73, opened 2019-07-09
Add a CSS variable to allow the d2l-navigation shadow to be hidden

AllanKerr:
# Changes

This pull request adds the `--d2l-navigation-shadow-drop-border-display` CSS variable to allow the shadow to be hidden on specific pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ Then add the `d2l-navigation`, and provide sub elements `d2l-navigation-main-hea
 </d2l-navigation>
 ```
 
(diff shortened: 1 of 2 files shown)


svanherk:
I think this is only needed for one page, similar to what was needed for the old gradient in grades.  Exposing the attribute would mean the navbar MVC renderer would need to be updated with an option to configure that, and that page would need to specify the special navbar type.  Still do-able, but not as clean as an override in css.  But you're right, this allows people more flexibility than we need

PR #74, opened 2019-07-24
Jroach/us108637

jeroach:
[US108637](https://rally1.rallydev.com/#/detail/userstory/322122660856?fdp=true): [UI] New web component that wraps d2l-navigation-iterator and treats the buttons as links

Thoughts on the name? `d2l-navigation-link-iterator` was my initial idea.

Also, since this is simply a logic wrapper, I don't think we need a designer sign off. Please let me know if you think otherwise

### DIFF
--- a/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
@@ -55,10 +55,14 @@ class D2LNavigationLinkIterator extends PolymerElement {
 
 	_onNavigationIteratorButtonClicked(evt) {
 		if (evt.detail.type === 'previous') {
-			window.location.href = this.previousHref;
+			this._setWindowLocationHref(this.previousHref);
 		} else {
-			window.location.href = this.nextHref;
+			this._setWindowLocationHref(this.nextHref);
 		}
+	}
+
+	_setWindowLocationHref(href) {
+		window.location.href = href;
 	}
 }
 
--- a/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
@@ -52,6 +52,7 @@ class D2LNavigationLinkIterator extends PolymerElement {
 				no-next=[[noNext]]
 				on-d2l-navigation-iterator-button-clicked="_onNavigationIteratorButtonClicked"
 			>
+			<slot></slot>
 			</d2l-navigation-iterator>
 		`;
 	}
--- a/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
@@ -55,9 +55,9 @@ class D2LNavigationLinkIterator extends PolymerElement {
 
 	_onNavigationIteratorButtonClicked(evt) {
 		if (evt.detail.type === 'previous') {
-			window.location.replace(this.previousHref);
+			window.location.href = this.previousHref;
 		} else {
-			window.location.replace(this.nextHref);
+			window.location.href = this.nextHref;
 		}
 	}
 }
--- a/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
@@ -1,5 +1,4 @@
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
-import 'd2l-typography/d2l-typography-shared-styles.js';
 import './d2l-navigation-iterator.js';
 
 /**
@@ -66,4 +65,4 @@ class D2LNavigationLinkIterator extends PolymerElement {
 	}
 }
 
-window.customElements.define('d2l-navigation-link-iterator', D2LNavigationLinkIterator);
+window.customElements.define(D2LNavigationLinkIterator.is, D2LNavigationLinkIterator);
--- a/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
@@ -1,0 +1,68 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
+import './d2l-navigation-iterator.js';
+
+/**
+ * @customElement
+ * @polymer
+ */
+class D2LNavigationLinkIterator extends PolymerElement {
+	static get is() { return 'd2l-navigation-link-iterator'; }
+
+	static get properties() {
+		return {
+			previousHref: {
+				type: String
+			},
+			nextHref: {
+				type: String
+			},
+			previousText: {
+				type: String
+			},
+			nextText: {
+				type: String
+			},
+			hideText: {
+				type: Boolean,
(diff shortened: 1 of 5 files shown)


components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
```diff
@@ -0,0 +1,69 @@
...
+				no-previous=[[noPrevious]]
+				no-next=[[noNext]]
+				on-d2l-navigation-iterator-button-clicked="_onNavigationIteratorButtonClicked"
+			>
+			<slot></slot>
+			</d2l-navigation-iterator>
+		`;
+	}
+
+	_onNavigationIteratorButtonClicked(evt) {
```
jeroach: Not sure how best to test this - from my brief digging, there does not seem to be an easy way to mock `window.location.replace`

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
```diff
@@ -0,0 +1,69 @@
...
+				type: Boolean,
+				value: false
+			}
+		};
+	}
+
+	static get template() {
+		return html`
+			<style>
+				:host {
```
jeroach: Not sure if I should just delete this, or leave empty.

CarolBusman: I don't think we need a `style` at all in a component - I'd say remove it.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
CarolBusman: My first thought would be `sinon`, but not sure that qualifies as "easy" considering our last experience with it.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
dlockhart: By default this will be `display: inline`, which is usually not what we want. So usually we at least have `display: inline-block;` here.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
```diff
@@ -0,0 +1,69 @@
...
+			</d2l-navigation-iterator>
+		`;
+	}
+
+	_onNavigationIteratorButtonClicked(evt) {
+		if (evt.detail.type === 'previous') {
+			window.location.replace(this.previousHref);
+		} else {
+			window.location.replace(this.nextHref);
+		}
```
dlockhart: By replacing the window's location, the back/forward buttons in the browser aren't going to work. Is that the behaviour you want? If not, this should be doing `window.location.href = whatever`.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: Thanks Dave, I'll put it back

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: Nope :) Thank you! Will update

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
dlockhart: I agree with Carol here -- I'd use sinon to mock out the `window` object and assert that the location is being set to the expected value when the iterator is clicked.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: Hmmm, ideally I just want to use the style of the child element - by setting `display: inline-block;` then it loses the flex layout of the child.

What's the best practice? Duplicate the style logic in the parent, or is there some way to just use the child styling when the parent is a wrapper?

dlockhart: Hmm, so in that case if just the default inline styling works, go for it.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: Doing some digging, I decided to make a helper method to set the `window.location.href` and then mock and verify the helper.

Rationale: https://stackoverflow.com/questions/4792281/mocking-window-location-href-in-javascript/7050884

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
```diff
@@ -0,0 +1,69 @@
...
+			this._setWindowLocationHref(this.nextHref);
+		}
+	}
+
+	_setWindowLocationHref(href) {
+		window.location.href = href;
+	}
+}
+
+window.customElements.define('d2l-navigation-link-iterator', D2LNavigationLinkIterator);
```
AlexBedley: nit: use `D2LNavigationLinkIterator.is`

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
```diff
@@ -0,0 +1,69 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
```
AlexBedley: is this import necessary?

AlexBedley:
`d2l-navigation-link-iterator` sounds like a fine name. Out of curiosity, how come you didn't just add this functionality to `d2l-navigation-iterator`?

Although there are no visual changes, there are still "UI" changes due to the behaviour of clicking a link. For example, they might have something to say about the back and forward browser buttons. But in this case I think it's fine to bypass a designer check.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: Nope :)

jeroach:
> `d2l-navigation-link-iterator` sounds like a fine name. Out of curiosity, how come you didn't just add this functionality to `d2l-navigation-iterator`?

The intent of `d2l-navigation-iterator` is to be generic (and fire events on click) instead of redirecting - this is to support it being used on a page that might:
a) use partial reloads (and not redirect to a page)
b) show and hide info via the iterator
c) other similar use case not involving a redirect

Thus the user can wrap `d2l-navigation-iterator` and do what they wish with the event.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: Copy paste thing

AlexBedley:
> > `d2l-navigation-link-iterator` sounds like a fine name. Out of curiosity, how come you didn't just add this functionality to `d2l-navigation-iterator`?
> 
> The intent of `d2l-navigation-iterator` is to be generic (and fire events on click) instead of redirecting - this is to support it being used on a page that might:
> a) use partial reloads (and not redirect to a page)
> b) show and hide info via the iterator
> c) other similar use case not involving a redirect
> 
> Thus the user can wrap `d2l-navigation-iterator` and do what they wish with the event.

Makes total sense, thanks.

jeroach:
Tests are timing out, mostly seems to be an issue with `IE11` :(

Must be related to `sinon` - I'm digging

PR #75, opened 2019-07-30
US108899 - Add on-click logic for previous and next 

jeroach:
[US108899](https://rally1.rallydev.com/#/detail/userstory/324245903592?fdp=true): [UI] Add two onClick properties (one for previous, one for next) to d2l-navigation-iterator

### DIFF
--- a/components/d2l-navigation-iterator/d2l-navigation-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-iterator.js
@@ -28,6 +28,12 @@ class D2LNavigationIterator extends PolymerElement {
 			noNext: {
 				type: Boolean,
 				value: false
+			},
+			previousOnClick: {
+				type: Object
+			},
+			nextOnClick: {
+				type: Object
 			}
 		};
 	}
@@ -67,6 +73,13 @@ class D2LNavigationIterator extends PolymerElement {
 
 	_dispatchButtonClicked(e) {
 		e.stopPropagation();
+
+		if (this.previousOnClick && e.currentTarget.type === 'previous') {
+			this.previousOnClick();
+		} else if (this.nextOnClick && e.currentTarget.type === 'next') {
+			this.nextOnClick();
+		}
+
 		this.dispatchEvent(new CustomEvent('d2l-navigation-iterator-button-clicked', {
 			detail: {
 				type: e.currentTarget.type
--- a/components/d2l-navigation-iterator/d2l-navigation-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-iterator.js
@@ -68,7 +68,7 @@ class D2LNavigationIterator extends PolymerElement {
 	_dispatchPreviousClicked(e) {
 		e.stopPropagation();
 
-		this.dispatchEvent(new CustomEvent('previous-clicked', {
+		this.dispatchEvent(new CustomEvent('previous-click', {
 			detail: {
 				type: e.currentTarget.type
 			}
@@ -78,7 +78,7 @@ class D2LNavigationIterator extends PolymerElement {
 	_dispatchNextClicked(e) {
 		e.stopPropagation();
 
-		this.dispatchEvent(new CustomEvent('next-clicked', {
+		this.dispatchEvent(new CustomEvent('next-click', {
 			detail: {
 				type: e.currentTarget.type
 			}
--- a/components/d2l-navigation-iterator/d2l-navigation-iterator.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-iterator.js
@@ -28,12 +28,6 @@ class D2LNavigationIterator extends PolymerElement {
 			noNext: {
 				type: Boolean,
 				value: false
-			},
-			previousOnClick: {
-				type: Object
-			},
-			nextOnClick: {
-				type: Object
 			}
 		};
 	}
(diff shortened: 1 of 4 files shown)


components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -28,6 +28,12 @@ class D2LNavigationIterator extends PolymerElement {
 			noNext: {
 				type: Boolean,
 				value: false
+			},
+			previousOnClick: {
```
jeroach: Method defined here

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -67,6 +73,13 @@ class D2LNavigationIterator extends PolymerElement {
 
 	_dispatchButtonClicked(e) {
 		e.stopPropagation();
+
+		if (this.previousOnClick && e.currentTarget.type === 'previous') {
+			this.previousOnClick();
```
jeroach: Method called here

demo/d2l-navigation-iterator.html
```diff
@@ -12,13 +12,21 @@
...
+
 		<script type="module">
 			import '@polymer/iron-demo-helpers/demo-pages-shared-styles';
 			import '@polymer/iron-demo-helpers/demo-snippet';
-			import 'd2l-typography/d2l-typography.js';
+			import 'd2l-typography/d2l-typography.js';
 			import '../d2l-navigation-separator.js';
 			import '../components/d2l-navigation-iterator/d2l-navigation-iterator.js';
 
+			function demoAlert() {
```
jeroach: Simple alert for demo purpose

demo/d2l-navigation-iterator.html
```diff
@@ -43,40 +51,46 @@ <h3>d2l-navigation-iterator empty demo</h3>
...
+			<h3>d2l-navigation-iterator no previous demo</h3>
+			<demo-snippet>
+				<template strip-whitespace>
+					<d2l-navigation-iterator no-previous></d2l-navigation-iterator>
+				</template>
+			</demo-snippet>
+			<h3>d2l-navigation-iterator on click demo</h3>
+			<demo-snippet>
+				<template strip-whitespace>
+					<d2l-navigation-iterator previous-on-click=demoAlert></d2l-navigation-iterator>
```
jeroach: What do I need to set this attribute to??

dlockhart:
Have a read-through of the [docs on events](https://polymer-library.polymer-project.org/3.0/docs/devguide/events#annotated-listeners). I think instead of these being properties (`previousOnClick`) you want these to just be custom events that get fired (`previous-click`).

That way other web components can just do:
`<d2l-navigation-iterator on-previous-click="handlePrevious">`

Things outside web components will need to register a listener more formally:
`elem.addEventListener('previous-click', function() { });`

demo/d2l-navigation-iterator.html
```diff
@@ -15,7 +15,7 @@
 		<script type="module">
 			import '@polymer/iron-demo-helpers/demo-pages-shared-styles';
 			import '@polymer/iron-demo-helpers/demo-snippet';
-			import 'd2l-typography/d2l-typography.js';
+			import 'd2l-typography/d2l-typography.js';
```
jeroach: Converted entire file to tabs - was mixed before

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -54,20 +54,33 @@ class D2LNavigationIterator extends PolymerElement {
...
 		e.stopPropagation();
-		this.dispatchEvent(new CustomEvent('d2l-navigation-iterator-button-clicked', {
+
+		this.dispatchEvent(new CustomEvent('d2l-navigation-iterator-previous-clicked', {
+			detail: {
+				type: e.currentTarget.type
+			},
+			bubbles: true,
+			composed: true
+		}));
```
dlockhart: I don't know that it necessarily needs to bubble or be composed? Do folks beyond the immediate parent need to know about it? The nice thing about not bubbling and not being composed is that you could call the event `previous-clicked` without worrying about namespacing. @dbatiste thoughts here?

jeroach: Hmmm, I guess only the immediate parent needs to know.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
```diff
@@ -45,7 +45,8 @@ class D2LNavigationLinkIterator extends PolymerElement {
 				hide-text=[[hideText]]
 				no-previous=[[noPrevious]]
 				no-next=[[noNext]]
-				on-d2l-navigation-iterator-button-clicked="_onNavigationIteratorButtonClicked"
+				on-d2l-navigation-iterator-previous-clicked="_onNavigationIteratorButtonClicked"
+				on-d2l-navigation-iterator-next-clicked="_onNavigationIteratorButtonClicked"
```
jeroach: To confirm: I've defined a handler method `_onNavigationIteratorButtonClicked` that does a `window.location.href = ...`

Can the legacy LMS page can still add its own event listener? What determines the order?

dlockhart: I don't know that the browser guarantees order... but in theory it would be whoever registers first.

components/d2l-navigation-iterator/d2l-navigation-iterator.js
dbatiste: Agreed.  Even if it needed to bubble in this context, if it's not required to bubble across the shadow-boundary then at least the `composed` bit can be turned off, which helps to limit the scope of the event.

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -54,25 +54,34 @@ class D2LNavigationIterator extends PolymerElement {
...
+				on-click="_dispatchNextClicked"></d2l-navigation-iterator-item>
 		`;
 	}
 
-	_dispatchButtonClicked(e) {
+	_dispatchPreviousClicked(e) {
 		e.stopPropagation();
-		this.dispatchEvent(new CustomEvent('d2l-navigation-iterator-button-clicked', {
+
+		this.dispatchEvent(new CustomEvent('previous-clicked', {
```
dlockhart: Just a slight nit on the name of the event here: I might go with `previous-click` so that the wire-up is `on-previous-click`. It just matches the tense of the native `onclick` (and `onfocus`, `onblur`, etc.) a little better.

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: @amauryperez and I chatted offline about this - in the case where additional logic is required when the event is fired, one should use the base `d2l-navigation-iterator` (and have the logic also handle the redirect after the other logic is done)

If you just want the redirect to happen, then the `d2l-navigation-link-iterator` works as intended

components/d2l-navigation-iterator/d2l-navigation-iterator.js
jeroach: Good call

components/d2l-navigation-iterator/d2l-navigation-link-iterator.js
jeroach: This way we don't need to worry about the timing of event handlers

PR #77, opened 2019-08-14
Jroach/us109273

jeroach:
[US109273](https://rally1.rallydev.com/#/detail/userstory/327823974944?fdp=true): [UI] Immersive Nav - Iterator Polish

Part 1 of 2
- This updates the component to support the desired `disabled` functionality of the iterator

Part 2 (what I'm working on now) will be on the LMS side, and will include:
- Updating to use the new iterator logic in this PR
- Fix spacing/padding/centering which currently isn't working as desired (component seems fine in isolation - issue crops up when it's used in the LMS)

**Updated disabled iterator state**

iterator_disabled_on_hover

**Updated disabled button**

updated_button_on_hover

### DIFF
--- a/components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
@@ -14,6 +14,15 @@ class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
 
 	static get properties() {
 		return {
+			disabled: {
+				type: Boolean,
+				value: false
+			},
+			hideText: {
+				type: Boolean,
+				value: false,
+				reflectToAttribute: true
+			},
 			text: {
 				type: String,
 				value: ''
@@ -22,11 +31,6 @@ class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
 				type: String,
 				value: 'previous',
 				observer: '_typeChanged',
-				reflectToAttribute: true
-			},
-			hideText: {
-				type: Boolean,
-				value: false,
 				reflectToAttribute: true
 			},
 			_icon: {
@@ -44,8 +48,11 @@ class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
 					height: 100%;
 					position: relative;
 				}
-				:host([hidden]) {
-					display: none;
+				:host([disabled]) {
+					color: var(--d2l-color-ferrite);
+					cursor: auto;
+					opacity: .5;
+					text-decoration: none;
 				}
 				:host([type="previous"]) d2l-navigation-button {
 					padding-left: 1.2rem;
@@ -84,13 +91,15 @@ class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
 					padding-right: 0;
 				}
 			</style>
-			<d2l-navigation-button id="d2l-navigation-iterator-item" class="d2l-navigation-iterator-item-button" text="[[_computeText(text, type)]]">
+			<d2l-navigation-button id="d2l-navigation-iterator-item" class="d2l-navigation-iterator-item-button" text="[[_computeText(text, type)]]" disabled="[[disabled]]">
 				<div class="d2l-navigation-iterator-item-items">
 					<d2l-icon icon$="[[_icon]]"></d2l-icon>
 					<span class="d2l-navigation-iterator-item-text" hidden="[[hideText]]">[[_computeText(text, type)]]</span>
 				</div>
 			</d2l-navigation-button>
-			<d2l-tooltip for="d2l-navigation-iterator-item">[[_computeText(text, type)]]</d2l-tooltip>
+			<template is="dom-if" if="[[!disabled]]">
+				<d2l-tooltip for="d2l-navigation-iterator-item">[[_computeText(text, type)]]</d2l-tooltip>
+			</template>
 		`;
 	}
 
--- a/components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
+++ b/components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
@@ -95,7 +95,7 @@ class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
 					padding-right: 0;
 				}
 			</style>
-			<d2l-navigation-button id="d2l-navigation-iterator-item" class="d2l-navigation-iterator-item-button" text="[[_computeText(text, type)]]" disabled$="[[disabled]]">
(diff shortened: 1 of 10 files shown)


components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -9,32 +9,36 @@ import '../../d2l-navigation-button.js';
...
-				computed: '_computeIcon(type)'
-			}
-		};
-	}
+class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
+	static get is() { return 'd2l-navigation-iterator-item'; }
+
+	static get properties() {
+		return {
+			disabled: {
```
jeroach: Updated to alpha order while I was renaming stuff

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -53,14 +53,14 @@ class D2LNavigationIterator extends PolymerElement {
 				text=[[previousText]]
 				type="previous"
 				hide-text=[[hideText]]
-				hidden=[[noPrevious]]
+				disabled=[[previousDisabled]]
 				on-click="_dispatchPreviousClicked"></d2l-navigation-iterator-item>
```
svanherk: I think these `on-click`'s should check if the button is disabled, and not fire an event if so

jeroach: Good call - thanks!

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -9,32 +9,36 @@ import '../../d2l-navigation-button.js';
...
+				type: Boolean,
+				value: false
```
dlockhart: Since you're using an attribute selector in CSS, this is a case where you'll want `reflectToAttribute`. Otherwise if folks set this in JavaScript `elem.disabled = true`, the attribute won't get set and your selector won't hit.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -84,13 +91,15 @@ class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
...
+			<d2l-navigation-button id="d2l-navigation-iterator-item" class="d2l-navigation-iterator-item-button" text="[[_computeText(text, type)]]" disabled="[[disabled]]">
 				<div class="d2l-navigation-iterator-item-items">
 					<d2l-icon icon$="[[_icon]]"></d2l-icon>
 					<span class="d2l-navigation-iterator-item-text" hidden="[[hideText]]">[[_computeText(text, type)]]</span>
 				</div>
 			</d2l-navigation-button>
-			<d2l-tooltip for="d2l-navigation-iterator-item">[[_computeText(text, type)]]</d2l-tooltip>
+			<template is="dom-if" if="[[!disabled]]">
+				<d2l-tooltip for="d2l-navigation-iterator-item">[[_computeText(text, type)]]</d2l-tooltip>
+			</template>
```
dlockhart: An alternate way to do this that would avoid ugly `dom-if` would be to use an attribute selector to hide the tooltip in CSS:

```css
:host([disabled]) d2l-tooltip {
    display: none;
}
```

components/d2l-navigation-iterator/d2l-navigation-iterator.js
dlockhart: Before adding that... double-check that it's required. If the `disabled` attribute gets passed down properly to a native `<button>` element, clicking on it won't actually fire a `click` event.

svanherk: Good point - might be worth adding a test to check that and ensure it keeps happening

jeroach: Event was actually firing since the element is our custom component (that wraps a button and tooltip).

I've updated the code to ensure the event is not fired if the button is disabled, and added a test to ensure this.

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
jeroach: Fixed, thanks! Much cleaner!

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
jeroach: Fixed, thanks! I find this logic bit confusing. I'll get it eventually :P

components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -84,7 +95,7 @@ class D2LNavigationIteratorItem extends NavigationLocalize(PolymerElement) {
 					padding-right: 0;
 				}
 			</style>
-			<d2l-navigation-button id="d2l-navigation-iterator-item" class="d2l-navigation-iterator-item-button" text="[[_computeText(text, type)]]">
+			<d2l-navigation-button id="d2l-navigation-iterator-item" class="d2l-navigation-iterator-item-button" text="[[_computeText(text, type)]]" disabled$="[[disabled]]">
```
jeroach: @dlockhart I assumed the `$` is needed now that I'm using `reflectToAttribute` - if not, please let me know and I'll remove it.

dlockhart: You shouldn't need it, no. You only need the `$` if you explicitly want to bind this to the attribute and not the property. In this case, binding to the property should be fine.

test/d2l-navigation-iterator.html
```diff
@@ -11,51 +11,72 @@
...
-		iter.noPrevious = !iter.noPrevious;
-		assert.equal(iter.noPrevious, prev.hidden);
-
-		assert.equal(iter.noNext, next.hidden);
-		iter.noNext = !iter.noNext;
-		assert.equal(iter.noNext, next.hidden);
-	});
-});
-</script>
+			import '@polymer/iron-test-helpers/mock-interactions.js';
```
jeroach: I fixed the formatting while I was in here

test/d2l-navigation-iterator.html
```diff
@@ -11,51 +11,72 @@
...
+					assert.equal(iter.previousDisabled, prev.disabled);
+					iter.previousDisabled = !iter.previousDisabled;
+					assert.equal(iter.previousDisabled, prev.disabled);
+
+					assert.equal(iter.nextDisabled, next.disabled);
+					iter.nextDisabled = !iter.nextDisabled;
+					assert.equal(iter.nextDisabled, next.disabled);
+				});
+				test('onClick events do not fire when buttons are disabled', function() {
+					iter = fixture('previousDisabled');
```
jeroach: fixture name is wrong here - fixing

PR #82, opened 2019-09-26
US109311 - Add Custom Scrollbars to Navigation Band for Org Tabs

svanherk:
- Add custom scrollbar styles to the content inside the slot of the navigation band
- Only do so if we detect we are on a non-mobile Windows machine (want to leave the nice mobile and mac scrollbars alone)

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,8 +174,11 @@ Then add the `d2l-navigation-band`.
 <d2l-navigation-band></d2l-navigation-band>
 ```
 
+The `d2l-navigation-band` also includes a `slot` with a custom scrollbar and fading effects, but this has only been designed for the `d2l-organization-consortium-tabs` and should not be used for anything else right now.
+
 ***Relevant CSS class name:***
 * `--d2l-branding-primary-color`: Used to customize the colour of the top navigation band.
+* `--d2l-navigation-band-slot-height`: When using the slot, this is needed to setup the proper scrollbar and fading effects.
 
 ---
 
--- a/d2l-navigation-band.js
+++ b/d2l-navigation-band.js
@@ -11,6 +11,15 @@ Polymer-based web component for a solid colour band that runs along the top of t
 */
 class D2LNavigationBand extends PolymerElement {
 
+	static get properties() {
+		return {
+			customScroll: {
+				type: Boolean,
+				value: false,
+				reflectToAttribute: true
+			}
+		};
+	}
 	static get template() {
 		const template = html`
 		${navigationSharedStyle}
@@ -19,6 +28,40 @@ class D2LNavigationBand extends PolymerElement {
 				background-color: var(--d2l-branding-primary-color, var(--d2l-color-celestine));
 				display: block;
 				min-height: 4px;
+
+				overflow-x: auto;
+				overflow-y: hidden;
+			}
+
+			:host([custom-scroll]) {
+				 /* Firefox Styles */
+				 scrollbar-color: var(--d2l-color-galena) var(--d2l-color-sylvite);
+				 scrollbar-width: thin;
+
+				 /* IE Styles */
+				 scrollbar-face-color:var(--d2l-color-galena);
+				 scrollbar-arrow-color: var(--d2l-color-sylvite);
+				 scrollbar-track-color: var(--d2l-color-sylvite);
+				 scrollbar-shadow-color: var(--d2l-color-sylvite);
+			 }
+			/* Webkit Styles */
+			:host([custom-scroll])::-webkit-scrollbar {
+				background-color: var(--d2l-color-sylvite);
+				height: 9px;
+			}
+			:host([custom-scroll])::-webkit-scrollbar-thumb {
+				background-color: var(--d2l-color-galena);
+				border-radius: 8px;
+				border-bottom: 1px solid var(--d2l-color-sylvite);
+				border-top: 1px solid var(--d2l-color-sylvite);
+			}
+			/* Styles to ensure the right padding is respected when scrolling */
+			.d2l-navigation-centerer {
+				line-height: 0;
+			}
+			.d2l-navigation-gutters {
+				display: inline-block;
+				line-height: 0;
 			}
 		</style>
 		<div class="d2l-navigation-centerer">
@@ -29,6 +72,13 @@ class D2LNavigationBand extends PolymerElement {
 		`;
 		template.setAttribute('strip-whitespace', '');
 		return template;
+	}
+	ready() {
+		super.ready();
+		const userAgent = navigator.userAgent.toLowerCase();
+		if (userAgent.indexOf('win') > -1 && userAgent.indexOf('mobile') === -1) {
+			this.customScroll = true;
+		}
 	}
 }
 customElements.define('d2l-navigation-band', D2LNavigationBand);
--- a/d2l-navigation-band.js
+++ b/d2l-navigation-band.js
@@ -15,8 +15,11 @@ class D2LNavigationBand extends PolymerElement {
 		return {
 			customScroll: {
 				type: Boolean,
-				value: false,
-				reflectToAttribute: true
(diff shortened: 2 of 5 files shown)


d2l-navigation-band.js
```diff
@@ -11,6 +11,15 @@ Polymer-based web component for a solid colour band that runs along the top of t
 */
 class D2LNavigationBand extends PolymerElement {
 
+	static get properties() {
+		return {
+			customScroll: {
```
svanherk: Wasn't sure if I should make this `_customScroll`.  I don't really intend for the clients to set this, but we don't usually reflect private variables to the attribute.  I guess I could make this private and `reflectToAttribute: false`, and set a different attribute name.  But all of this doesn't stop the user from setting this attribute themselves, it just makes it less inviting to do so.  Thoughts?

d2l-navigation-band.js
```diff
@@ -19,6 +28,40 @@ class D2LNavigationBand extends PolymerElement {
...
+				background-color: var(--d2l-color-sylvite);
+				height: 9px;
+			}
+			:host([custom-scroll])::-webkit-scrollbar-thumb {
+				background-color: var(--d2l-color-galena);
+				border-radius: 8px;
+				border-bottom: 1px solid var(--d2l-color-sylvite);
+				border-top: 1px solid var(--d2l-color-sylvite);
+			}
+			/* Styles to ensure the right padding is respected when scrolling */
```
svanherk: Had to add the below after _a lot_ of investigating around the right padding/margin just disappearing completely when using native scrollbars (https://www.brunildo.org/test/scroll-child-margin.html)

d2l-navigation-band.js
dlockhart: In Lit what I think we'd do is calculate this in the `render()` method and then use it to apply a CSS class to a `<div>` (instead of on the host). That way nothing leaks out in the form of a property or attribute on the custom element itself.

I'm not sure you can do that in Polymer, since the `template()` method is static.

d2l-navigation-band.js
```diff
@@ -19,6 +28,40 @@ class D2LNavigationBand extends PolymerElement {
...
+				overflow-y: hidden;
+			}
+
+			:host([custom-scroll]) {
+				 /* Firefox Styles */
+				 scrollbar-color: var(--d2l-color-galena) var(--d2l-color-sylvite);
+				 scrollbar-width: thin;
+
+				 /* IE Styles */
+				 scrollbar-face-color:var(--d2l-color-galena);
```
dlockhart: Nit: missing space

d2l-navigation-band.js
```diff
@@ -11,6 +11,15 @@ Polymer-based web component for a solid colour band that runs along the top of t
 */
 class D2LNavigationBand extends PolymerElement {
 
+	static get properties() {
+		return {
+			customScroll: {
+				type: Boolean,
+				value: false,
```
dlockhart: I think you can avoid the bit in the constructor by using a function to define the value. Also you could make it readOnly to avoid folks being able to set it:

```
customScroll: {
  type: Boolean,
  readOnly: true,
  value: function() {
    const userAgent = navigator.userAgent.toLowerCase();
    if (userAgent.indexOf('win') > -1 && userAgent.indexOf('mobile') === -1) {
      this.customScroll = true;
    }
  }
}
```

test/navigation-band.html
```diff
@@ -15,13 +15,61 @@
...
+					userAgent: 'Windows Mac OS X',
+					result: true
+				}].forEach(function({ userAgent, result }) {
+					setup(function() {
+						originalUserAgent = navigator.userAgent;
+					});
+					teardown(function() {
+						Object.defineProperty(window.navigator, 'userAgent', {value: originalUserAgent, configurable: true});
+					});
+					test('custom scrollbars are ' + (result ? 'on ' : 'off ') + 'for user agent of "' + userAgent + '"', function() {
```
dlockhart: Nit: template literals would make this much more readable:

    test(`custom scrollbars are ${result ? 'on' : 'off'} for user agent of "${userAgent}"

jeroach:
Looks good on the demo locally :)
scrollbar_in_band

d2l-navigation-band.js
jeroach: I like the "avoid constructor" suggestion Dave had below.

As to the two first options (private vs public and reflect or not) I'm indifferent.

d2l-navigation-band.js
svanherk: I like this!  I'll also try applying the styles to the div instead of the host, so it's in a closer state to what we'd want for Lit to make that migration smoother.

test/navigation-band.html
svanherk: Will update, for some reason I thought this wouldn't work for IE11 tests but the polyfill must be adding this

d2l-navigation-shared-styles.js
```diff
@@ -16,18 +16,33 @@ export const navigationSharedStyle = html`
 				position: relative;
 			}
 
+			.d2l-navigation-scroll[custom-scroll]:before,
```
svanherk: This feels kinda weird in this file and I can move it to `d2l-navigation-band.js`, but I liked it living here so if these media query values or padding values ever change, someone would know to update these too.

d2l-navigation-band.js
```diff
@@ -25,60 +28,77 @@ class D2LNavigationBand extends PolymerElement {
...
 				height: 9px;
 			}
-			:host([custom-scroll])::-webkit-scrollbar-thumb {
+			.d2l-navigation-scroll[custom-scroll]::-webkit-scrollbar-thumb {
 				background-color: var(--d2l-color-galena);
 				border-radius: 8px;
 				border-bottom: 1px solid var(--d2l-color-sylvite);
 				border-top: 1px solid var(--d2l-color-sylvite);
 			}
+			/* Faded edges styles */
```
svanherk: Jeff requested a fading effect at the edges to make it obvious when you have more tabs

README.md
```diff
@@ -174,8 +174,11 @@ Then add the `d2l-navigation-band`.
 <d2l-navigation-band></d2l-navigation-band>
 ```
 
+The `d2l-navigation-band` also includes a `slot` with a custom scrollbar and fading effects, but this has only been designed for the `d2l-organization-consortium-tabs` and should not be used for anything else right now.
+
 ***Relevant CSS class name:***
 * `--d2l-branding-primary-color`: Used to customize the colour of the top navigation band.
+* `--d2l-navigation-band-slot-height`: When using the slot, this is needed to setup the proper scrollbar and fading effects.
```
svanherk: I tried measuring this value based on the size of the slot's content, but it quickly made this "simple" component very complicated, with all the issues we've seen with `ResizeObserver` and running measuring code when most clients won't even see these tabs.  The org tabs are a set size (`1.5rem`), instead of just hardcoding that I did expose this variable to keep some of the customization ability, but this is definitely a trade-off - sacrificing flexibility for simplicity, at least while we have no plans for any other usages of this.

d2l-navigation-band.js
```diff
@@ -25,60 +28,77 @@ class D2LNavigationBand extends PolymerElement {
 		${navigationSharedStyle}
 		<style>
 			:host {
-				background-color: var(--d2l-branding-primary-color, var(--d2l-color-celestine));
+				background: linear-gradient(180deg, var(--d2l-branding-primary-color, var(--d2l-color-celestine)) var(--d2l-navigation-band-slot-height, 1.5rem), #ffffff 0%);
```
svanherk: This adds a white background against the scrollbar area, rather than having the colour overhang when the scrollbar doesn't cover the whole width of the screen.

svanherk:
image
image
image
Smaller screen:
image

^ Screenshots from demo page.  The real tabs look a lot nicer, don't worry 😝

PR #89, opened 2019-10-17
US111258 - Allow nav-band to listen for a scroll event

svanherk:
This is the first of three PRs adding the ability for the selected tab to be scrolled into view and centered.

Since we can't use `ScrollIntoView` because that moves the focus in Chrome, we need to tell the `d2l-navigation-scroll` element where we want to go.  But only the contents of the slot know where we want to go, so this PR adds the ability for the `d2l-navigation-band` to receive an event (`d2l-navigation-band-slot-scroll-request`) with the location of the point that the nav-band should try to center.

### DIFF
--- a/d2l-navigation-band.js
+++ b/d2l-navigation-band.js
@@ -122,8 +122,8 @@ class D2LNavigationBand extends PolymerElement {
 
 		var scrollAmount = e.detail.pointToCenter - 0.5 * scroll.offsetWidth;
 
-		if (dir === 'rtl') {
-			// Chrome
+		if (dir && dir.toLowerCase() === 'rtl') {
+			// Chrome and Safari
 			//scrollAmount = scroll.scrollWidth + e.detail.pointToCenter - 1.5 * scroll.offsetWidth;
 
 			//Firefox
@@ -132,7 +132,7 @@ class D2LNavigationBand extends PolymerElement {
 			//Edge
 			//scrollAmount = 0.5 * scroll.offsetWidth - e.detail.pointToCenter;
 		}
-		requestAnimationFrame(function() {
+		requestAnimationFrame(() => {
 			scroll.scrollLeft = scrollAmount;
 		});
 	}
--- a/d2l-navigation-band.js
+++ b/d2l-navigation-band.js
@@ -36,6 +36,7 @@ class D2LNavigationBand extends PolymerElement {
 			.d2l-navigation-scroll {
 				overflow-x: auto;
 				overflow-y: hidden;
+				scroll-behavior: smooth;
 			}
 
 			.d2l-navigation-scroll[custom-scroll] {
@@ -101,6 +102,24 @@ class D2LNavigationBand extends PolymerElement {
 		`;
 		template.setAttribute('strip-whitespace', '');
 		return template;
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.addEventListener('d2l-navigation-band-slot-scroll-request', this._handleScrollRequest);
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.removeEventListener('d2l-navigation-band-slot-scroll-request', this._handleScrollRequest);
+	}
+
+	_handleScrollRequest(e) {
+		e.stopPropagation();
+		const scroll = this.shadowRoot.querySelector('.d2l-navigation-scroll');
+		requestAnimationFrame(function() {
+			scroll.scrollLeft = e.detail.pointToCenter - (scroll.offsetWidth / 2);
+		});
 	}
 }
 customElements.define('d2l-navigation-band', D2LNavigationBand);
--- a/d2l-navigation-band.js
+++ b/d2l-navigation-band.js
@@ -31,6 +31,7 @@ class D2LNavigationBand extends PolymerElement {
 				background: linear-gradient(180deg, var(--d2l-branding-primary-color, var(--d2l-color-celestine)) var(--d2l-navigation-band-slot-height, 1.5rem), #ffffff 0%);
 				display: block;
 				min-height: 4px;
+				position: relative;
 			}
 
 			.d2l-navigation-scroll {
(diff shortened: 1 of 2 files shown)


d2l-navigation-band.js
```diff
@@ -36,6 +36,7 @@ class D2LNavigationBand extends PolymerElement {
 			.d2l-navigation-scroll {
 				overflow-x: auto;
 				overflow-y: hidden;
+				scroll-behavior: smooth;
```
svanherk: This is not supported in IE11, Edge or Safari, so in those browsers it will just jump to the new location.  Jeff was ok with this, since IE11 is going away and Chredge will support this.

d2l-navigation-band.js
```diff
@@ -102,5 +103,23 @@ class D2LNavigationBand extends PolymerElement {
...
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.removeEventListener('d2l-navigation-band-slot-scroll-request', this._handleScrollRequest);
+	}
+
+	_handleScrollRequest(e) {
+		e.stopPropagation();
+		const scroll = this.shadowRoot.querySelector('.d2l-navigation-scroll');
+		requestAnimationFrame(function() {
+			scroll.scrollLeft = e.detail.pointToCenter - (scroll.offsetWidth / 2);
```
svanherk: `scrollTo` is broken in Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/15534521/

demo/navigation.html
```diff
@@ -146,6 +146,27 @@ <h3>d2l-navigation - navigation band content slotted - scrollbar</h3>
 		</div>`;
 
 document.body.appendChild($_documentContainer.content);
+</script>
+<script type="module">
+	requestAnimationFrame(function() {
+		const scrollTestButtons = document.querySelectorAll('.scrollable');
+		scrollTestButtons.forEach(function(button) {
+			button.addEventListener('click', function() {
```
dlockhart: Nit: you can use arrow functions for all these callbacks

d2l-navigation-band.js
jeroach: How does this work for RTL? Does `scrollLeft` need to become `scrollRight` - or am I misunderstanding?

dlockhart: That's a good question actually... 🤔

svanherk: `scrollRight` doesn't exist, but I was under the impression that the browser would handle automatically, or that the values returned would switch to negative and I'd have to play around with the +/-.  But neither is happening... the values returned seem to be fairly random.  I'm still trying to figure it out, but there's very little documented about it and no consistency.  I found this, but it just kinda ends... https://readable-email.org/list/www-style/topic/cssom-view-value-of-scrollleft-in-rtl-situations-is-completely-busted-across-browsers

svanherk: Going to try t check if this is still accurate, it's pretty old now: https://github.com/othree/jquery.rtl-scroll-type

d2l-navigation-band.js
```diff
@@ -102,5 +104,37 @@ class D2LNavigationBand extends PolymerElement {
...
+	}
+
+	_handleScrollRequest(e) {
+		e.stopPropagation();
+		const dir = document.getElementsByTagName('html')[0].getAttribute('dir');
+		const scroll = this.shadowRoot.querySelector('.d2l-navigation-scroll');
+
+		var scrollAmount = e.detail.pointToCenter - 0.5 * scroll.offsetWidth;
+
+		if (dir === 'rtl') {
```
svanherk: @jeroach @dbatiste @dlockhart @ctwomey1 @geurts 

Finally getting back to this after all the conference craziness!  This solution is a bit involved, so I wanted to go back and absolutely confirm `<element>.scrollIntoView()` wouldn't work for us.  I walked away from it originally because it causes Chrome to "move the focus" to that element (the `activeElement` does not change, but the next time the user presses tab the focus will go to the tabbable element that comes after that scrolled-to element in the DOM).  I tried to fix this by calling `scrollIntoView` back on the `activeElement`, but the timing of this is super finicky (calling `scrollIntoView` again will cancel the old one currently in progress).  And on top of that, the issue still remains that calling this will scroll the element into view horizontally AND vertically. Which means that, if a user is taken back to their previous location lower down on a page after a refresh or pressing the back button, their view will be scrolled to the top so we can see the tabs scroll, then moved back down to where they were... not great.

So `ScrollIntoView` is out.  The solution I have here and in https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/74 and https://github.com/BrightspaceHypermediaComponents/organizations/pull/129 remains if we're ok with it.  I would just need to apply the changes code reviewers have made, and figure out the RTL stuff. https://github.com/othree/jquery.rtl-scroll-type ended up being very helpful... every browser implements `scrollLeft` in RTL differently.  I've figured out the different calculations (which you can see below) for each browser except for Safari, which I'm _hoping_ will match Chrome.  We can turn off the feature for "unknown browsers", but having to do browser detection at all here feels weird... thoughts?

I'm going to outline our options below, and wanted to get dev and design feedback for what we want:

1) Don't scroll the tab into view, if we don't find it's worth the code we'll need to add or are super concerned about it.
2) Use the solution I have here, and add the browser detection for RTL.
3) Use the solution I have here, but instead _turn off_ the scroll-tab-into-view feature for RTL.
4) Redesign the feature to not include automatic scrolling (for example, reorder the tabs such that the selected tab is always first, followed by the rest in alphabetical order)

I would estimate all of these are about the same amount, other than the first one of "do nothing" haha.  I'm going to figure out how to make a gif and throw that in this thread as well, to try to showcase what this PR's experience feels like.

svanherk: scrolling-no-cache

geurts: This is where it would be great to know if we have any user using RTL locale's in a consortium :P

I know we avoid browser detection wherever possible, so I'll leave it up to you to determine how icky or acceptable that is. It wouldn't be the end of the world if we had to go with option # 1, since it *mostly* only affects admin users; however it does affect users enrolled in just a few schools if they're using this in the mobile side menu. For this reason, I'd rather go with option # 3 since I'm sure we'd be affecting very few users (if any).

I can't think of a way to put the active tab first without negatively impacting the UX (muscle memory etc). I'm open to other ideas though.

d2l-navigation-band.js
```diff
@@ -102,5 +104,37 @@ class D2LNavigationBand extends PolymerElement {
...
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.removeEventListener('d2l-navigation-band-slot-scroll-request', this._handleScrollRequest);
+	}
+
+	_handleScrollRequest(e) {
+		e.stopPropagation();
+		const dir = document.getElementsByTagName('html')[0].getAttribute('dir');
```
dlockhart: Looking at the `RtlMixin`, it does things slightly differently:

    const dir = document.documentElement.getAttribute('dir') || 'ltr';

They're probably equivalent though?

d2l-navigation-band.js
dlockhart: Yeah this seems like a case where the complexity of the code -- and likelihood for it to break in the future out of nowhere or become a maintenance nightmare -- outweighs RTL support. If we can stomach it, I'd treat it as a known limitation.

svanherk: I agree, I think #3 is our best option and I confirmed with Sandra that she's good with this.  I could definitely see this changing in the future with no warning, hopefully for the better if all browsers agree on how this should _actually_ work.

d2l-navigation-band.js
svanherk: I can use that instead to be consistent!

svanherk:
Throwing this here in case we do get a request for this someday.  To get this to work in RTL, these are the calculations you'd need to use:

```
// Chrome and Safari
scrollAmount = scroll.scrollWidth + e.detail.pointToCenter - 1.5 * scroll.offsetWidth;

// Firefox
scrollAmount = e.detail.pointToCenter - 0.5 * scroll.offsetWidth;

// Edge
scrollAmount = 0.5 * scroll.offsetWidth - e.detail.pointToCenter;
```

PR #91, opened 2019-12-03
Remove tabstop on navigation-band in Firefox

svanherk:
When a div is scrollable, Firefox makes it a tab stop by default.  We don't want/need that functionality, so turning that off here.


svanherk:
This worked in the demo page, but when I add it to the LMS with the real org tabs it doesn't fix it (seems to then just place it on another div) 😞 Oh well!

PR #92, opened 2020-01-07
De35352/Alt attribute on image with no link/text

Scy-D2L:
- Added an index for the list of demo pages
- When passed an empty text value or missing text prop, `d2l-navigation-link-image` now creates inner `` tag with `alt` attribute.
- Added test to confirm the above
- Refactored the button tests
- Tested code changes with BSI on a local lms instance

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -1,0 +1,21 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
+		<title>d2l-navigation demo list</title>
+	</head>
+	<body>
+		<h1>Navigation Demo Pages</h1>
+		<ul>
+			<li><a href="navigation.html">navigation</a></li>
+			<li><a href="navigation-band.html">navigation-band</a></li>
+			<li><a href="d2l-navigation-button.html">navigation-button</a></li>
+			<li><a href="navigation-immersive.html">navigation-immersive</a></li>
+			<li><a href="d2l-navigation-iterator.html">navigation-iterator</a></li>
+			<li><a href="d2l-navigation-link-iterator.html">navigation-link-iterator</a></li>
+			<li><a href="navigation-main-footer.html">navigation-main-footer</a></li>
+			<li><a href="navigation-main-header.html">navigation-main-header</a></li>
+			<li><a href="navigation-separator.html">navigation-separator</a></li>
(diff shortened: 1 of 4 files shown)


demo/index.html
```diff
@@ -0,0 +1,21 @@
...
+			<li><a href="d2l-navigation-button.html">navigation-button</a></li>
+			<li><a href="navigation-immersive.html">navigation-immersive</a></li>
+			<li><a href="d2l-navigation-iterator.html">navigation-iterator</a></li>
+			<li><a href="d2l-navigation-link-iterator.html">navigation-link-iterator</a></li>
+			<li><a href="navigation-main-footer.html">navigation-main-footer</a></li>
+			<li><a href="navigation-main-header.html">navigation-main-header</a></li>
+			<li><a href="navigation-separator.html">navigation-separator</a></li>
+		</ul>
+	</body>
+</html>
```
dlockhart: Nice!

test/d2l-navigation-button.html
```diff
@@ -258,6 +263,18 @@
...
 	});
 });
+
+suite('d2l-navigation-link-image undefined text', function() {
+	setup(function() {
+		button = fixture('imageLinkUndefinedText');
+	});
+	testInstantiation('d2l-navigation-link-image');
+	test('should have empty alt when text attribute is missing or an empty string', function() {
+		var img = dom(button.root).querySelector('img');
```
dlockhart: I know all the other tests do it this way, but can you try `button.shadowRoot.querySelector`? That's the Polymer 3 / Lit way of querying children in the shadow DOM that we'd like to transition to.

test/d2l-navigation-button.html
```diff
@@ -258,6 +263,18 @@
 		assert.equal(button.text, text);
 	});
 });
+
+suite('d2l-navigation-link-image undefined text', function() {
```
dlockhart: Despite other stuff in here not doing it, feel free to start using arrow functions (`suite('blah', () => { ... })`) and other ES6 things like `const` and `let` instead of `var`.

test/d2l-navigation-button.html
Scy-D2L: One better: just updated the whole file. Verified it's all still working ;)

PR #94, opened 2020-03-02
Fix nav bar visible-on-ancestor for the Polymer to Lit dropdown conversion

AllanKerr:
# Issue
In the nav bar, the `visible-on-ancestor` functionality broke because the Lit version loads faster than the Polymer version. This means that when the `visible-on-ancestor-mixin` uses [`findComposedAncestor`](https://github.com/BrightspaceUI/core/blob/master/mixins/visible-on-ancestor-mixin.js#L104), the parent element has loaded but its shadow DOM has not. Because the `d2l-visible-on-ancestor-target` is hidden within the ancestor's shadow DOM, the look up fails.

# Fix
To fix this, `d2l-visible-on-ancestor-target` needs to be
moved outside of the ancestor's shadow DOM into the light DOM.

### DIFF
--- a/d2l-navigation-main-footer.js
+++ b/d2l-navigation-main-footer.js
@@ -20,7 +20,7 @@ class D2LNavigationMainFooter extends PolymerElement {
 					display: block;
 				}
 			</style>
-			<div class="d2l-navigation-centerer d2l-visible-on-ancestor-target">
+			<div class="d2l-navigation-centerer">
 				<div class="d2l-navigation-gutters">
 					<slot name="main"></slot>
 				</div>
--- a/test/navigation-main-footer.html
+++ b/test/navigation-main-footer.html
@@ -26,10 +26,6 @@ suite('d2l-navigation-main-footer', function() {
 		expect(centerer).to.not.be.null;
 		assert.equal(centerer.tagName, 'DIV');
 
-		var ancestor = dom(footer.root).querySelector('.d2l-visible-on-ancestor-target');
-		expect(ancestor).to.not.be.null;
-		assert.equal(ancestor.tagName, 'DIV');
-
 		var gutters = dom(footer.root).querySelector('.d2l-navigation-gutters');
 		expect(gutters).to.not.be.null;
 		assert.equal(gutters.tagName, 'DIV');


AllanKerr:
> Wouldn't this be a problem if visible-on-ancestor was used in another web component? (ex. [d2l-card](https://github.com/BrightspaceUI/card/blob/master/d2l-card.js#L166) defines itself as visible-on-ancestor target, and even if it were moved to it's parent, it too is in a shadowDOM).
> 
> Perhaps we can make `__voaInit` smarter... perhaps by waiting a microtask, or waiting a frame? We need to be careful to avoid any flash... i.e. we don't want to show it and then hide it.
> 
> I think @margaree ran into a similar problem with hierarchical-view, when it tries to determine if it's a nested view.

I talked to @margaree about it and her issue only occured in Edge while this seems to happen in Chrome and Firefox.

I haven't been able to figure out why the nav bar case is special but it seems to work properly with `d2l-card` since the card's shadow DOM is loaded when `__voaInit` is called. I'll look into that case a bit more to see if I can figure out why the loading order seems to be different.

dbatiste:
Not sure how it plays into things in this case, but `d2l-card` is still a Polymer element - I'd expect it to be even slower.  🤔

AllanKerr:
> Not sure how it plays into things in this case, but `d2l-card` is still a Polymer element - I'd expect it to be even slower. 🤔

The distinction between the two seems to be that, with the nav bar, the `connectedCallback` is triggered immediately after the page `onload` event is fired. With `d2l-card`, the `connectedCallback` is triggered during a later `render` call at which point the ancestor's shadow DOM has already loaded.

I can look into making adding a microtask wait to the `visible-on-ancestor-mixin` to see if that fixes it.

AllanKerr:
> 
> I can look into making adding a microtask wait to the `visible-on-ancestor-mixin` to see if that fixes it.

Discussed this with @svanherk yesterday and we're agreeance that this seems to be a better option. Everything I've found suggests that rendering shadow DOM can be `async` meaning we can't rely on the shadow DOM being loaded at the time the mixin's `connectedCallback` is called.

PR #95, opened 2020-04-08
Support a new `width-type` attribute

wongvincent:
The user may now apply a `width-type` attribute on `d2l-navigation-immersive` with the value `normal`.
`normal` restricts the nav bar left/middle/right to fit a max-width of `1230px`.

The existing CSS variable, `--d2l-navigation-immersive-content-max-width` which allowed the user to configure the width to be any value will be removed. Usages of the existing CSS variable (2 places) will be updated accordingly to use the new attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Include the [webcomponents.js](http://webcomponents.org/polyfills/) "lite" polyf
 Then add the `d2l-navigation-immersive`, providing values for the `backLinkHref` & `backLinkText`. Additionally, you may override any of the 3 slots (`left`, `middle`, `right`).
 Please note that overridding the `left` slot will prevent the Back link from displaying. This should only be done in very specialized cases.
 
-`d2l-navigation-immersive` can optionally have its max-width constrained to `1230px` by including the attribute `width-type="normal"`
+Optionally, the max-width can be configured to match the max-width used by the LE by setting `widthType` to `normal`.
 
 <!---
 ```
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Include the [webcomponents.js](http://webcomponents.org/polyfills/) "lite" polyf
 Then add the `d2l-navigation-immersive`, providing values for the `backLinkHref` & `backLinkText`. Additionally, you may override any of the 3 slots (`left`, `middle`, `right`).
 Please note that overridding the `left` slot will prevent the Back link from displaying. This should only be done in very specialized cases.
 
-`d2l-navigation-immersive` can optionally have a max width set to match your content.  Simply set the `--d2l-navigation-immersive-content-max-width` css variable to your desired width.
+`d2l-navigation-immersive` can optionally have its max-width constrained to `1230px` by including the attribute `width-type="normal"`
 
 <!---
 ```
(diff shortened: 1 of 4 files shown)


d2l-navigation-immersive.js
```diff
@@ -248,5 +257,18 @@ class D2LNavigationImmsersive extends PolymerElement {
...
 	}
+
+	_widthTypeChanged(widthType) {
+		if (widthType) {
+			const elem = dom(this.root).querySelector('.d2l-navigation-immersive-container');
+			switch (this.widthType) {
+				case 'normal':
+					elem.style.maxWidth = '1230px';
+					break;
+				case 'fullscreen':
```
jeroach, line 268: Should this `switch` not have a `default` case?

svanherk:
I'm a bit confused... since the variable exists, can your component/app not set that?  Or are you thinking we should keep the value of `1230px` constrained to this repo as much as possible?

If so, I still think we should set the variable instead of overriding it... maybe a Boolean property that reflects to attribute instead, and hooks into css selectors?

dlockhart:
Basically we don't want a million places all knowing about the `1230` value -- we want to be able to change the width of our pages in the future (as we've done many times in the past) without going into tons of different places.

To resolve it, I suggested to Vincent that we introduce a `width-type` to our page templates (we just have one right now) and to the nav components that can either be `normal` or `fulllscreen`. Normal would restrict it to the `1230px` value, but we can change that in the future easily.

@wongvincent, I think there's actually [only 1 usage](https://search.d2l.dev/search?full=%22--d2l-navigation-immersive-content-max-width%22&project=lms) of the variable, and unsurprisingly it's setting it to `1230`. So we could clean all of this up and just get rid of it altogether. :)

I also think the code solution is even easier... you could do it with as simple CSS selector:

```css
.d2l-navigation-immersive-container {
    /* existing CSS */
    max-width: 100%;
}
:host([width-type="normal"]) .d2l-navigation-immersive-container {
    max-width: 1230px;
}
```

You'll need to make sure the attribute is reflected though for that to work.

svanherk:
Ok cool, discussed more with Dave during standup and also @geurts.  I'm good with cleaning up the css variable altogether, and Jeff is going to update the design.d2l immersive nav docs to remove the comment about being able to set any width.

Looks like it's also used in `folio-app` (https://search.d2l.dev/xref/GitHub/Brightspace/folio-app/src/components/nav-bar/nav-bar.js?r=908b16f9#11), but it's an FRA so we could fix that after merge if needed.

README.md
```diff
@@ -91,7 +91,7 @@ Include the [webcomponents.js](http://webcomponents.org/polyfills/) "lite" polyf
 Then add the `d2l-navigation-immersive`, providing values for the `backLinkHref` & `backLinkText`. Additionally, you may override any of the 3 slots (`left`, `middle`, `right`).
 Please note that overridding the `left` slot will prevent the Back link from displaying. This should only be done in very specialized cases.
 
-`d2l-navigation-immersive` can optionally have a max width set to match your content.  Simply set the `--d2l-navigation-immersive-content-max-width` css variable to your desired width.
+`d2l-navigation-immersive` can optionally have its max-width constrained to `1230px` by including the attribute `width-type="normal"`
```
svanherk, line 94: Minor: Should we be more generic here and say "can optionally have its max-width set to match the max-width used by the LE", so we don't need to update this if we ever change that value?

wongvincent:
Merged with `[increment minor]`, but it didn't create a new release? 😕

PR #96, opened 2020-04-13
deploying as a separate stage

dlockhart:
This switches the deployment to be a separate stage, which allows us to avoid the check to only run it in the visual-diff stage. It also lets us restrict the encrypted variables to the stages for which they're required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ jobs:
         echo "Not a pull request and/or no secure environment variables, running headless tests...";
         npm run test:polymer:local || travis_terminate 1;
       fi
+    env:
+      - SAUCE_USERNAME: Desire2Learn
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,31 +17,33 @@ jobs:
...
   - stage: visual-difference-tests
     script:
     - |
       if [ $TRAVIS_SECURE_ENV_VARS == true ]; then
         echo "Running visual difference tests...";
         npm run test:diff || travis_terminate 1;
       fi
-after_success:
-- |
-  if [ $TRAVIS_BUILD_STAGE_NAME == "Visual-difference-tests" ]; then
```
dlockhart, line 29: I think it was the uppercase "V" here that was causing this to never get run. Not sure how it ever worked, but maybe a change to how Travis works.

PR #97, opened 2020-07-14
restoring back-link-href and back-link-text properties

dlockhart:
Not sure how this happened, but these 2 properties were [accidentally deleted](https://github.com/BrightspaceUI/navigation/commit/74425b5736614d97a39e91cfc3322ab3ae4b5df0#diff-be945901e9cdfaab1afb2753c6bb090a) last July. Restoring them.

### DIFF
--- a/d2l-navigation-immersive.js
+++ b/d2l-navigation-immersive.js
@@ -18,6 +18,14 @@ class D2LNavigationImmsersive extends PolymerElement {
 
 	static get properties() {
 		return {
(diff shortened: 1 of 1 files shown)


bemailloux:
`Out of curiosity` Does the version need to change whenever a UI component is modified?

dlockhart:
Yes. When this merges, I'll increment the patch version, then BSI will need to pull in that version for it to appear in the LMS.

PR #99, opened 2020-09-29
prevent back link text & icon from wrapping

dlockhart:
Noticed that when things get tight, the back-link can wrap both inside the text but also the text can wrap on a different line from the back chevron icon. Since these are always used inside the navigation header, I don't think we ever want wrapping.


jeroach (approved):

svanherk (approved):
Makes sense, I don't know if anyone was relying on this wrapping for small screens but like you said I don't think we ever _want_ it to wrap.  Soon we'll be adding the "shortening" feature as well, where at small screens it changes the text to just "Back" instead of "Back to ___"

PR #100, opened 2020-10-01
increasing z-index on immersive nav to be higher than skeleton loaders

dlockhart:
@Vitalii-Misechko noticed that when immersive nav (`z-index: 1`) is used with skeleton loaders (`z-index: 1`), when you scroll the page such that the skeletons are "behind" the immersive nav, whichever rendered last "wins" the z-index tie and is displayed on top:

image (1)

Increasing immersive nav to `z-index: 2` solves the problem. The band also uses `z-index: 2` for the scrolling gradient, but as far as I can tell bands with stuff inside them that might scroll and immersive nav can't be used together. Does that sound right Stacey?

### DIFF
--- a/d2l-navigation-immersive.js
+++ b/d2l-navigation-immersive.js
@@ -47,7 +47,7 @@ class D2LNavigationImmsersive extends PolymerElement {
 				position: fixed;
 				right: 0;
 				top: 0;
-				z-index: 1;
+				z-index: 2; /* higher than skeletons which could scroll behind immersive nav */
 			}
 			d2l-navigation {
 				border-bottom: 1px solid var(--d2l-color-mica);


jeroach (approved):
Sounds reasonable to me

svanherk:
> as far as I can tell bands with stuff inside them that might scroll and immersive nav can't be used together. Does that sound right Stacey?

Currently yes, we only needed it for org tabs in the main LMS navbar so we never exposed that ability to the immersive nav.  We never really discussed whether we'd want that eventually, but I feel it's unlikely.  Even if a number of our workflows keep updating to use the immersive nav, the whole point is for it to be small - I can't see us wanting to add org tabs or something else to it.  FYI @geurts in case he has heard otherwise.

Either way, we could address that when we go back to add the ability for immersive nav to include items in the nav-band slot

svanherk (approved):

PR #101, opened 2020-10-21
upgrading to puppeteer 5

dlockhart:
`visual-diff` has puppeteer as a peer dependency, which gets installed as of NPM 7 and is conflicting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 cache:
   npm: false
 language: node_js
-node_js: node
+node_js:
+- "14"
 addons:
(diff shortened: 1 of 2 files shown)


AlexBedley (approved):
Change looks good, not sure why the build is failing though. Heads up that @jeroach is on vacation this week

svanherk:
This is the error the `facet-filter-sort` repo saw with node 15

dlockhart:
Alright, I've locked it at Node 14 and it's working. So that's what we'll have to do for Polymer repos for now.

PR #102, opened 2020-10-23
d2l-navigation-button - Add property to remove top bar

amauryperez:
image

This is the default behavior on the iterator when hovering over the button. See the blue bar at the top

image

This is what design has asked for when the button is used not at the top of the page. No blue bar at the top


d2l-navigation-button.js
```diff
@@ -107,14 +112,17 @@ class D2LNavigationButton extends mixinBehaviors([D2L.PolymerBehaviors.Button.Be
 					}
 				</style>
 			<button aria-describedby$="[[_ariaDescriptionId]]" aria-expanded$="[[ariaExpanded]]" aria-haspopup$="[[ariaHaspopup]]" aria-label$="[[text]]" class="d2l-focusable" disabled$="[[disabled]]" autofocus$="[[autofocus]]" form$="[[form]]" formaction$="[[formaction]]" formenctype$="[[formenctype]]" formmethod$="[[formmethod]]" formnovalidate$="[[formnovalidate]]" formtarget$="[[formtarget]]" name$="[[name]]" title$="[[text]]" type$="[[type]]">
-			<span class="d2l-navigation-button-top-border"></span>
+			<span class$="[[_getTopBorderClass()]]" class="d2l-navigation-button-top-border"></span>
```
dbatiste, line 115: This seems odd to be - I think `class` should only be specified once on the element.

dbatiste (commented):
Looks good... just the `class` duplication to deal with.

d2l-navigation-button.js
amauryperez: yeah, my bad, forgot to remove that second one

dbatiste (approved):

dlockhart:
The other option here would be to introduce a new type of button in core -- a circular button or something. But this is good for now.

PR #105, opened 2020-11-26
Adding a simple workflow to test the setup of the AWS role and token rotator

svanherk:
This will be deleted after I see everything is working, but I can't test this in my fork.  This tests whether I can assume the role successfully with the rotated secrets and then lists out the first item in the bucket (which should be the screenshots folder).


dlockhart (approved):
Good luck!

PR #107, opened 2020-12-10
ci: remove visual-diff from CI workflow

dlockhart:
This was done as an experiment to see how the timing compared to having it in its separate `visual-diff.yml` workflow. We've decided we like having the separate workflow better as it allows the tests to run in parallel and they can also be re-runnable.


svanherk (approved):
Thanks!  Just added the visual diff check as required *was waiting until we decided).

jeroach (approved):

PR #109, opened 2020-12-11
Change default colour

svanherk:
Let things be red now!


svanherk:
Tricked ya!  This was just for demo purposes.

PR #111, opened 2021-01-04
Enabled immersive nav view to handle RTL

dustxd:
### Motivation/Context
During FACE development, when trying to add a title to the immersive navigation bar, we noticed that the immersive navigation bar doesn't take into account RTL as the divider between the "Back to {Activity}" button and "Title" renders properly in LTR but disappears in RTL languages. The reason is because we're setting `border-right: none` on `.d2l-navigation-immersive-middle` regardless of language.

#### Current Behaviour
_Notice the divider is rendered on the left_

image

#### Expected Behaviour
image

### Summary of Changes
* Added mixin behaviour to `d2l-navigation-immersive` so that it inherits the `dir` attribute with value equal to either `ltr` or `rtl`
* Added CSS selector to apply the right styling to the host when the locale is a `rtl` language

### Rally Story
https://rally1.rallydev.com/#/29180338367d/dashboard?detail=%2Fuserstory%2F461506956220&fdp=true?fdp=true

### Related PR
https://github.com/Brightspace/lms/pull/5084

### DIFF
--- a/d2l-navigation-immersive.js
+++ b/d2l-navigation-immersive.js
@@ -7,6 +7,7 @@ import './d2l-navigation-link-back.js';
 import { navigationSharedStyle } from './d2l-navigation-shared-styles.js';
 import { dom } from '@polymer/polymer/lib/legacy/polymer.dom.js';
 import { PolymerElement, html } from '@polymer/polymer/polymer-element.js';
+import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
 
 /**
 `d2l-navigation-immersive`
@@ -14,7 +15,7 @@ Polymer-based web component for the immersive navigation component
 
 @demo demo/navigation-immersive.html
 */
-class D2LNavigationImmsersive extends PolymerElement {
+class D2LNavigationImmsersive extends mixinBehaviors([D2L.PolymerBehaviors.FocusableBehavior], PolymerElement) {
 
 	static get properties() {
 		return {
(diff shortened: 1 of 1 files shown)


d2l-navigation-immersive.js
```diff
@@ -7,14 +7,15 @@ import './d2l-navigation-link-back.js';
...
+import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
 
 /**
 `d2l-navigation-immersive`
 Polymer-based web component for the immersive navigation component
 
 @demo demo/navigation-immersive.html
 */
-class D2LNavigationImmsersive extends PolymerElement {
+class D2LNavigationImmsersive extends mixinBehaviors([D2L.PolymerBehaviors.FocusableBehavior], PolymerElement) {
```
dustxd, line 18: Is this an appropriate mixin to use? Or is there a better choice?

dlockhart: I had to look it up (it's been a while since I was in Polymer-land!), but you want the [DirMixin](https://polymer-library.polymer-project.org/3.0/api/mixins/dir-mixin). I've switched this to use that.

dlockhart (approved):

d2l-navigation-immersive.js
svanherk: Thanks Dave!  Was also trying to find it, we got it for free a lot by chance with what we included.

svanherk (approved):
When you merge, you'll want to add `fix: ` to the beginning of the merge commit to get an automatic patch release.  Thanks!

dustxd:
Thanks for the reminder! @svanherk

PR #117, opened 2021-11-11
increase z-index for the immersive navigation

JikaiLong:
This PR is to solve a visual problem [TA159188](https://rally1.rallydev.com/#/?detail=/task/614091004409&fdp=true): minor: Quiz items: settings page causes data reload

the current tooltip has a `z-index` of 1001 which means it is going to be on top of the immersive nav component, increasing the immersive nav `z-index` slightly could fix this visual problem.

### DIFF
--- a/d2l-navigation-immersive.js
+++ b/d2l-navigation-immersive.js
@@ -48,7 +48,7 @@ class D2LNavigationImmsersive extends DirMixin(PolymerElement) {
 				position: fixed;
 				right: 0;
 				top: 0;
-				z-index: 2; /* higher than skeletons which could scroll behind immersive nav */
+				z-index: 1002; /* higher than skeletons which could scroll behind immersive nav */
 			}
 			d2l-navigation {
 				border-bottom: 1px solid var(--d2l-color-mica);


svanherk:
Wouldn't we want a tooltip to go over the immersive nav so the user can read it?  Similarly with dropdowns?

jeroach:
> Wouldn't we want a tooltip to go over the immersive nav so the user can read it? Similarly with dropdowns?

Yeah I agree. I can't exactly remember the specifics but we've been burned in the past by playing around with `z-index`. If we can solve this another way that would be safer imho, but again it's been a while.

svanherk:
> Yeah I agree. I can't exactly remember the specifics but we've been burned in the past by playing around with z-index. If we can solve this another way that would be safer imho, but again it's been a while.

Yeah z-index is so fragile.  It'd help to know the exact scenario, but I think with tooltip you can define a boundary so you could let it know the immersive nav is above it and not to open there, or tell it to always open down if you don't want it opening up and covering the immersive nav.

JikaiLong:
This is a problem we are currently getting, 
Once we hover over a section with tooltips, the section will be brought forward as the same z-index as the tooltip which covers the navigation. 
image

dbatiste:
Does this change the stacking for dialogs or dropdowns in that context?  Their `z-index` start at 1000, and I think that when interacting with them, we would expect them to be on top of the nav.  To test this, it may be necessary to resize the browser such that the dialog would overlap the nav, or the dropdown would open upward.

svanherk:
> Once we hover over a section with tooltips, the section will be brought forward as the same z-index as the tooltip which covers the navigation.

Whoa, that is unexpected...  tooltip is somehow doing that? 😕 @dbatiste does that seem right?

svanherk:
Going to close this due to staleness, but feel free tor reopen if it's still an issue and there's some time to investigate Dave B's points

PR #119, opened 2021-12-14
feat: switch to use `core` for several components

devpow112:
The ones removed were all drop in replacements. The ones not removed used various style shims or polymer behaviours that can't be removed at the moment. Doing this as part of an effort to minimize github/git style dependency refs in package.json files. NPM 8 has issues with them and can greatly increase install time (`d2l-rubric` takes 2 min to install with NPM 6 and without any changes takes 30 min with NPM 8).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @jeroach @svanherk @CarolBusman
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @jeroach @svanherk @CarolBusman
+* @BrightspaceUI/gaudi-dev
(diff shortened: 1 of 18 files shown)


components/d2l-navigation-iterator/d2l-navigation-iterator-item.js
```diff
@@ -1,131 +1,131 @@
...
-		return (type === 'previous') ? this.localize('previous') : this.localize('next');
-	}
-
-	_computeIcon(type) {
-		return (type === 'previous') ? 'd2l-tier3:chevron-left-circle' : 'd2l-tier3:chevron-right-circle';
-	}
-}
-
-window.customElements.define('d2l-navigation-iterator-item', D2LNavigationIteratorItem);
+import '@brightspace-ui/core/components/colors/colors.js';
```
devpow112, line 1: This looks like an EOL chance to LF from CRLF. Hiding white-space diff shows the changes in this file correctly.

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @jeroach @svanherk @CarolBusman
```
devpow112, line 1: Should this just be updated to Gaudi?

dlockhart: At this point, probably yeah.

dlockhart (approved):

PR #120, opened 2021-12-14
fix: add `github:` to applicable dependencies

devpow112:


dlockhart:
Does this help with the slowness?

dlockhart (approved):

devpow112:
> Does this help with the slowness?

A bit but super minor. Just saves it trying to figure out which endpoint to go to. NPM supports raw `git`, `gitlab` and `github`. Overall not a large impact but it feels nicer to call this out explicitly.

PR #121, opened 2021-12-14
fix: add `components/lang` to `.npmignore`

devpow112:


dlockhart (approved):
I'm not sure if using the `files` entry in `package.json` works with things that don't publish to NPM, but I find it easier to explicitly say what you want vs. the other way around.

devpow112:
> 



> I'm not sure if using the `files` entry in `package.json` works with things that don't publish to NPM, but I find it easier to explicitly say what you want vs. the other way around.

Fair I can look into that. I've just been in the habit of using `.npmignore`.

PR #123, opened 2022-01-05
upgrading to use @web/dev-server and Lit demo components

dlockhart:
Actual plan: upgrade this repo from Polymer -> Lit

This seemed like a small piece of work that I could put into a PR on its own in the path to doing the Lit upgrade. This switches the demos over to use the new Lit demo components and to use the `@web/dev-server` instead of `polymer serve`.

Due to the Polymer 3 automated upgrade mess, it's going to be difficult to grok the changes, but I mostly kept everything as it was. I deleted a few demos because I felt they weren't really showing anything special.


demo/index.html
```diff
@@ -13,9 +13,6 @@ <h1>Navigation Demo Pages</h1>
 			<li><a href="navigation-immersive.html">navigation-immersive</a></li>
 			<li><a href="d2l-navigation-iterator.html">navigation-iterator</a></li>
 			<li><a href="d2l-navigation-link-iterator.html">navigation-link-iterator</a></li>
-			<li><a href="navigation-main-footer.html">navigation-main-footer</a></li>
-			<li><a href="navigation-main-header.html">navigation-main-header</a></li>
-			<li><a href="navigation-separator.html">navigation-separator</a></li>
```
dlockhart, line 18: I pulled the separator demo into `navigation.html` and didn't feel demoing the main header/footer added much since they're already used in `navigation.html`.

demo/navigation-immersive-demos/navigation-immersive-no-middle-slot.html
```diff
@@ -1,92 +0,0 @@
-<!doctype html>
```
dlockhart, line 1: I removed the demos where the middle/right slots were missing as I didn't feel they added much beyond the "all slots" demo.

svanherk: IIRC I think there's some separator hide/show logic when certain slots are missing?

demo/navigation-immersive.html
```diff
@@ -1,83 +1,28 @@
...
-	<body unresolved class="d2l-typography">
-		<script type="module">
-const $_documentContainer = document.createElement('template');
-
-$_documentContainer.innerHTML = `<div>
-			<h3>d2l-navigation-immersive</h3>
-			<iframe src="./navigation-immersive-demos/navigation-immersive-all-slots.html"></iframe>
-
-			<h3>d2l-navigation-immersive (consumer hides the right slot at 900px and the middle slot at 700px)</h3>
-			<iframe src="./navigation-immersive-demos/navigation-immersive-responsive-slot-hiding.html"></iframe>
```
svanherk, line 65: Did the responsive one maybe demonstrate the `ResizeObserver` behaviour of hiding/showing the separators?

README.md
```diff
@@ -237,6 +231,14 @@ npx mocha './test/**/*.visual-diff.js' -t 10000 -g some-pattern
 npx mocha './test/**/*.visual-diff.js' -t 10000 --golden
 ```
 
+### Running the demos
```
svanherk, line 234: Feels like this is in a weird spot?

dlockhart: I agree it's a bit weird, but I modelled it after what `npm init @brightspace-ui` output.

demo/navigation-immersive.html
dlockhart: It did yeah. Not sure what belong in a demo and what should actually be a visual diff test -- which I'm planning to bolster in a follow up PR. Just seemed like a lot of copy-paste across files with these immersive demos just to show a slightly different thing.

svanherk: Yeah I think a visual-diff test would be good enough. I just remember the slot empty detection being finicky so would be good to have a test for it at least.

svanherk (approved):

demo/navigation-immersive.html
dlockhart: For sure, I'll make sure I include it.

PR #125, opened 2022-01-06
Visual-Diff

dlockhart:
Adds visual-diff tests for all the components in preparation for the Lit migration.


.gitignore
```diff
@@ -1,4 +1,4 @@
-node_modules
+node_modules/
 package-lock.json
-/test/**/screenshots/current/
-/test/**/screenshots/golden/
+test/screenshots/current/
+test/screenshots/golden/
```
dlockhart, line 4: I moved the visual-diff tests to be inline with the rest of the tests, more like a typical repo.

demo/index.html
```diff
@@ -8,7 +8,6 @@
 		<h1>Navigation Demo Pages</h1>
 		<ul>
 			<li><a href="navigation.html">navigation</a></li>
-			<li><a href="navigation-band.html">navigation-band</a></li>
```
dlockhart, line 11: I merged the band demo into the main "navigation" one just like separator.

test/immersive.visual-diff.html
```diff
@@ -0,0 +1,39 @@
...
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta charset="UTF-8">
+		<script type="module">
+			import '@brightspace-ui/core/components/icons/icon.js';
+			import '@brightspace-ui/core/components/typography/typography.js';
+			import '../../d2l-navigation-immersive.js';
+		</script>
+	</head>
+	<body class="d2l-typography">
+		<template id="normal">
```
dlockhart, line 15: The immersive tests are slightly special in that they're all initially in a `<template>` so they don't all get stamped into the page. Pages are only meant to have one.

test/immersive.visual-diff.mjs
```diff
@@ -0,0 +1,86 @@
+/* eslint-disable no-invalid-this */
+import puppeteer from 'puppeteer';
+import { VisualDiff } from '@brightspace-ui/visual-diff';
+
+async function initTest(page, name) {
+	await page.$eval(`#${name}`, (template) => {
+		const clone = template.content.cloneNode(true);
+		document.body.insertBefore(clone, document.body.firstChild);
```
dlockhart, line 8: This does the stamping.

test/immersive.visual-diff.mjs
```diff
@@ -0,0 +1,86 @@
...
+	});
+	await page.waitForTimeout(100);
+}
+
+async function reset(page) {
+	await page.evaluate(() => {
+		const elem = document.querySelector('d2l-navigation-immersive');
+		elem.parentNode.removeChild(elem);
```
dlockhart, line 16: And this removes the cloned node after each test.

test/immersive.visual-diff.mjs
```diff
@@ -0,0 +1,86 @@
...
+					});
+					await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
+				});
+			});
+
+		});
+
+	});
+
+	['right', 'middle'].forEach((slot) => {
```
dlockhart, line 66: These are testing the resize observers that are watching the middle & right slots for new content to show/hide the border.

test/index.html
```diff
@@ -10,14 +10,14 @@
...
 			var testVariants = [], tests = [
 				'navigation.html',
-				'navigation-immersive.html',
-				'navigation-band.html',
-				'navigation-main-header.html',
-				'navigation-separator.html',
-				'd2l-navigation-button.html',
-				'navigation-main-footer.html',
-				'd2l-navigation-iterator.html',
-				'd2l-navigation-iterator-item.html'
```
dlockhart, line 20: I renamed these files to not have `d2l-navigation` in them.

test/iterator.visual-diff.html
```diff
@@ -0,0 +1,75 @@
...
+			import '@brightspace-ui/core/components/typography/typography.js';
+			import '../../components/d2l-navigation-iterator/d2l-navigation-iterator.js';
+			import '../../components/d2l-navigation-iterator/d2l-navigation-iterator-item.js';
+			import { forceFocusVisible } from '@brightspace-ui/core/helpers/focus.js';
+			window.forceFocusVisible = forceFocusVisible;
+		</script>
+		<style>
+			.wrapper {
+				display: inline-block;
+				padding: 0 50px 50px 50px;
```
dlockhart, line 19: This wrapper lets us snap a picture of the tooltip as well.

test/immersive.visual-diff.mjs
```diff
@@ -0,0 +1,94 @@
...
+			it(`${slot}`, async function() {
+				await initTest(page, `no-${slot}-slot`);
+				await page.evaluate((slot) => {
+					const newContent = document.createElement('div');
+					newContent.setAttribute('slot', slot);
+					newContent.innerHTML = 'New Content';
+					document.querySelector('d2l-navigation-immersive').appendChild(newContent);
+				}, slot);
+				await page.waitForTimeout(100);
+				await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
```
dlockhart, line 88: Thing I learned which was causing this to fail: I had the call to `reset(page)` right after this line. But when the visual diff fails, nothing after this line actually runs, which meant it wasn't getting reset which caused every other test to fail.

dlockhart: It wasn't failing locally because I was using the `--golden` flag as I was working on things which bypasses the failure.

svanherk:
One more case that I remember being super fragile - the hover/focus effect on the buttons overhangs and the header slots use negative margins to account for that, so the buttons line up with the content in the LMS.  I think if you remove something it doesn't overhang and instead the hover effect is what lines up with the content, the button shifted over.  And if you remove something else, the hover/focus effect is cutoff.  I think that's what the 7px stuff all over the place is.  Maybe just a focus test in both `navigation` and `navigation-immersive`?

dlockhart:
Ok I've added an immersive test that focuses on the back button. I don't see any negative margins anywhere else though -- nothing in navigation or the header stuff, unless I'm missing something.

I don't love a lot of the stuff in here... there's also code that relies on consumers using particular CSS classes on things which isn't great. I'll see what I can do to clean it up as either part of the migration or afterwards.

svanherk:
Ugh, found it... it's in BSI: https://search.d2l.dev/xref/Brightspace/brightspace-integration/sass/navigation/header.scss?r=d7677959#12

Yeah, not sure if it can be done in the migration or if it'll need to wait for "move to core", but there's a lot that could be simplified I think

dlockhart:
Oh wow. Yeah those are the CSS classes that consumers need to use and I want to clean up -- we'll see.

test/immersive.visual-diff.mjs
svanherk: Cool strategy!

test/index.html
svanherk: And soon this file will go away altogether 🎉

svanherk (approved):
Nice!

test/immersive.visual-diff.mjs
dlockhart: When you mentioned the fullscreen mobile dropdown issues you were having this morning in standup I thought of potentially doing something like this. This is also kind of what I want our future HTML-file-less visual-diff plans to involve... you set up a little fixture in your JS and it automatically stamps it into a document for you and then removes it later.

PR #129, opened 2022-01-10
Lit Part 1

dlockhart:
Just dipping my toes in here with some easy ones -- `<d2l-navigation>` and `<d2l-navigation-separator>`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,8 +20,6 @@ jobs:
           node-version-file: .nvmrc
       - name: Install dependencies
         run: npm install
-      - name: Unit Tests (Headless)
-        run: npm run test:polymer:local
       - name: Semantic Release
         uses: BrightspaceUI/actions/semantic-release@main
         with:
--- a/d2l-navigation.js
+++ b/d2l-navigation.js
@@ -1,22 +1,15 @@
-
-import '@polymer/polymer/polymer-legacy.js';
-
 import './d2l-navigation-band.js';
-import { navigationSharedStyle } from './d2l-navigation-shared-styles.js';
+import { css, html, LitElement } from 'lit-element';
 
-import { PolymerElement, html } from '@polymer/polymer/polymer-element.js';
 /**
-`d2l-navigation`
-Polymer-based web component for the all encompasing navigation wrapper component
+ * Primary navigation wrapper component.
+ * @slot - Default content placed inside of the component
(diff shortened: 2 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -20,8 +20,6 @@ jobs:
           node-version-file: .nvmrc
       - name: Install dependencies
         run: npm install
-      - name: Unit Tests (Headless)
-        run: npm run test:polymer:local
```
dlockhart, line 24: Missed this last time around. Don't see a great reason to redo the unit tests during the release workflow.

d2l-navigation.js
```diff
@@ -1,22 +1,15 @@
...
+ */
+class Navigation extends LitElement {
 
-@demo demo/navigation.html
-*/
-class D2LNavigation extends PolymerElement {
-
-	static get template() {
-		const template = html`
-		${navigationSharedStyle}
```
dlockhart, line 18: As far as I can tell, and as far as the testing I did could see, including these shared styles here was unnecessary. 

The header is the only place using the `--d2l-navigation-margin-regular` variable, and it also includes the shared styles so it'll get redefined there. Also nothing is actually setting this [except places that have copied the styles](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22--d2l-navigation-margin-regular%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5).

The rest of the stuff in the shared styles wasn't doing anything since none of those CSS classes are used by `<d2l-navigation>`. I tested any slotted content using those CSS classes wouldn't have been targeted by it anyway.

package.json
```diff
@@ -4,8 +4,10 @@
   "repository": "https://github.com/BrightspaceUI/navigation.git",
   "scripts": {
     "build": "gulp build",
-    "lint": "npm run lint:eslint",
+    "lint": "npm run lint:eslint && npm run lint:lit && npm run lint:style",
     "lint:eslint": "eslint . --ext .js,.html",
+    "lint:lit": "lit-analyzer d2l-navigation.js d2l-navigation-separator.js --strict --rules.no-unknown-tag-name off",
```
dlockhart, line 9: This will be annoying until everything is converted over.

svanherk (approved):
Looks good to me!

PR #131, opened 2022-02-10
sub old d2l-offscreen with core version

annabelsegalld2l:


d2l-navigation-button.js
```diff
@@ -115,7 +107,7 @@ class D2LNavigationButton extends mixinBehaviors([D2L.PolymerBehaviors.Button.Be
 			<span class$="[[_getTopBorderClass()]]"></span>
 			<slot></slot>
 			</button>
-			<span id="[[_ariaDescriptionId]]" class="d2l-offscreen-description">[[ariaDescribedbyText]]</span>
+			<d2l-offscreen id="[[_ariaDescriptionId]]" class="d2l-offscreen-description">[[ariaDescribedbyText]]</d2l-offscreen>
```
dlockhart, line 110: If nothing else is referencing it, I think the CSS class attribute can go away?

PR #133, opened 2022-04-05
Update README.md

pseudostack:
There doesn't appear to be a clear command to install this component, similar to what is avialable on https://github.com/BrightspaceUI/core.  Trying an NPM install of the same format that's available for CORE components doesn't work here so I've added the NPM install command that works for this navigation component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,10 @@
+## Installation
+
+```shell
+npm install BrightspaceUI/navigation
+```
+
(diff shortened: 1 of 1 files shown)


svanherk:
Looks like we missed this PR. Installation info was just added to the README in https://github.com/BrightspaceUILabs/navigation/pull/188

PR #138, opened 2022-06-14
CodeArtifact Switch Phase 2

devpow112:
This PR includes the following changes:
- [x] Switch all workflows to use Node 16
  - Replace the number in `.nvmrc` with `16`
- [x] Removes all GitHub references
  - Updates `package.json` to point to the new `NPM` packages
  - Updates all the import paths to use the new names


dlockhart:
Node 16 actually checks peer dependencies, so I think that's what's failing the build.

devpow112:
> Node 16 actually checks peer dependencies, so I think that's what's failing the build.

Yup looks like it just needed a eslint upgrade.

devpow112:
All format changes are automatic via `eslint --fix`.

PR #143, opened 2022-06-23
fix: make link behave like button

dlockhart:
The original intend was for `<d2l-navigation-link>` and `<d2l-navigation-button>` to be used interchangeably -- you just use whichever one matches your semantic needs. Unfortunately, [we discovered](https://stackoverflow.com/questions/15764600/what-makes-the-text-on-a-button-element-vertically-centered/34612505#34612505) some pretty weird behaviour that's internal to native `button`s, and that causes a difference in layout between links and buttons when the same CSS is applied.

The fix is to use a consistent `inline-block` on the outer web component, and `inline-flex` on the inner button/link. Additionally, `vertical-align: middle` brings all items on the same "line" into alignment vertically. This slight variation is going to result in a slight vertical shift in the visual diffs, and required a tweak to `d2l-navigation-link-back` which was doing something similar itself.

The only place using `<d2l-navigation-link>` directly is here in activity feed:
Screen Shot 2022-06-23 at 3 04 22 PM

Weirdly, they're not even setting an `href` so they don't get the "hover top border"... seems like not the right choice for them. Anyway, I applied these changes there and things still look fine.

`<d2l-navigation-link>` is also used from `<d2l-navigation-link-image>`, which is only used in the LMS navbar. It seems good with these changes as well.

Once these changes are in, I'm going to follow up with adding a `<d2l-navigation-link-icon>` and `<d2l-navigation-button-icon>` that specifically handles this "icon + text" case that keeps coming up. That way we can get consistent spacing between them. I also want to try and convert more of this to Lit and clean up some of the Polymer style mixins.


d2l-navigation-link-back.js
```diff
@@ -62,10 +62,6 @@ class D2LNavigationLinkBack extends mixinBehaviors([D2L.PolymerBehaviors.Focusab
 				padding-left: 0;
 				padding-right: var(--d2l-navigation-link-back-left-padding, 5px);
 			}
-			span, d2l-icon {
-				vertical-align: middle;
-				height: 100%;
-			}
```
dlockhart, line 68: No longer needed with the changes to `<d2l-navigation-link>`.

d2l-navigation-link.js
```diff
@@ -27,20 +27,24 @@ class D2LNavigationLink extends mixinBehaviors([D2L.PolymerBehaviors.FocusableBe
 			${highlightStyles}
 			<style is="custom-style">
 				:host {
-					display: flex;
+					display: inline-block;
 					height: 100%;
 				}
 				a {
-					@apply --d2l-navigation-highlight-base;
```
dlockhart, line 34: A lot of the CSS from this Polymer style mixin was for `<button>`s, not links (like margins/padding/etc.), so I copied it inline and removed the redundant stuff.

d2l-navigation-link.js
```diff
@@ -27,20 +27,24 @@ class D2LNavigationLink extends mixinBehaviors([D2L.PolymerBehaviors.FocusableBe
...
+					position: relative;
+					text-decoration: none;
+					vertical-align: middle;
 				}
 				a:hover,
 				a:focus {
-					@apply --d2l-navigation-highlight-base-hover-focus;
+					color: var(--d2l-color-celestine);
+					--d2l-icon-fill-color: var(--d2l-color-celestine);
+					outline: none;
```
dlockhart, line 47: Also copied this one inline in preparation for killing these style mixins.

d2l-navigation-link.js
```diff
@@ -49,11 +53,12 @@ class D2LNavigationLink extends mixinBehaviors([D2L.PolymerBehaviors.FocusableBe
 				.d2l-navigation-link-top-border {
 					@apply --d2l-navigation-highlight-border;
 				}
+				:host(:not([href])) .d2l-navigation-link-top-border {
+					display: none;
+				}
```
dlockhart, line 58: This was just an optimization to remove the need for the `<template is="dom-if">`.

d2l-navigation-link.js
```diff
@@ -27,20 +27,24 @@ class D2LNavigationLink extends mixinBehaviors([D2L.PolymerBehaviors.FocusableBe
...
-					display: flex;
+					display: inline-block;
 					height: 100%;
 				}
 				a {
-					@apply --d2l-navigation-highlight-base;
-					font-family: inherit;
-					outline: none; /* Needed for Edge, outline setting from css mixin is not being applied in Polymer 2 & 3 */
-				}
-				:host(:not([href])) a {
```
dlockhart, line 38: Also it was applying `cursor: pointer` to the link -- which links have by default anyway -- and then removing it if no `href` is specified... which links also do by default.

abcotter (approved):
Tested through Demo with adding my own links etc. 

LGTM :) thanks

PR #145, opened 2022-06-27
switch to use playwright for unit tests

dlockhart:


svanherk (approved):
Could also update the test path here if you want to match the template, but 🤷

dlockhart:
> Could also update the test path here if you want to match the template, but 🤷

I've got a bunch of subsequent PRs lined up after this one, so I'll do that in a follow-up.

PR #151, opened 2022-07-08
add d2l-navigation-button-icon component

dlockhart:
A new component that mirrors the functionality of the recently added `<d2l-navigation-link-icon>`, except with button semantics. Loosely based on the more generic `<d2l-navigation-button>` that takes a `<slot>`. I'm hoping to switch all usages of `<d2l-navigation-button>` over to this and get rid of it.

When the text is hidden, I have it rendering a tooltip instead of using the `title` -- this matches what the iterators are doing and will allow me to just switch them over to use this. I've updated link to do this as well, which isn't a breaking change since [the only usages of them](https://github.com/Brightspace/nova/blob/4f8d56b5459f90ad082a0b8004415171b410cdbc/src/app/app-header.js#L414) don't hide the text (FYI @abcotter).


README.md
```diff
@@ -253,11 +268,11 @@ Then run the tests:
 
 ```shell
 # run visual-diff tests
-npx mocha './test/**/*.visual-diff.mjs' -t 10000
+npx mocha './test/**/*.visual-diff.js' -t 10000
```
dlockhart, line 271: Missed this last time around.

d2l-navigation-button-icon.js
```diff
@@ -0,0 +1,112 @@
...
+			button:not([disabled]):hover,
+			button:not([disabled]):focus {
+				--d2l-icon-fill-color: var(--d2l-color-celestine);
+				color: var(--d2l-color-celestine);
+				outline: none;
+			}
+			button[disabled] {
+				cursor: default;
+				opacity: 0.5;
+			}
```
dlockhart, line 79: These styles are ported over from the `d2l-navigation-shared-styles.js` Polymer style mixins, but then heavily modified since there's no `<slot>`, no IE11/Edge hacks and the "highlight border" styles are already in a shared export.

d2l-navigation-button-icon.js
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	render() {
+		const ariaLabel = this.textHidden ? this.text : undefined;
+		const highlightBorder = !this.disabled ? html`<span class="d2l-navigation-highlight-border"></span>` : nothing;
+		const id = this.textHidden ? this._buttonId : undefined;
+		const text = !this.textHidden ? this.text : nothing;
+		const tooltip = this.textHidden ? html`<d2l-tooltip for="${this._buttonId}" for-type="label">${this.text}</d2l-tooltip>` : nothing;
+		return html`
+			<button id="${ifDefined(id)}" ?disabled="${this.disabled}" aria-label="${ifDefined(ariaLabel)}">
```
dlockhart, line 101: I'm setting `aria-label` AND associating the tooltip [because of this defect](https://github.com/BrightspaceUI/core/issues/2583).

test/button.test.js
```diff
@@ -1,10 +1,9 @@
 import '../d2l-navigation-button.js';
 import '../d2l-navigation-button-close.js';
+import '../d2l-navigation-button-icon.js';
 import '../d2l-navigation-button-notification-icon.js';
-import '../d2l-navigation-link.js';
-import '../d2l-navigation-link-back.js';
-import '../d2l-navigation-link-image.js';
```
dlockhart, line 6: I split the link tests out into `link.test.js` to match what the visual-diffs were doing.

test/button.visual-diff.html
```diff
@@ -25,6 +26,18 @@
 		<div class="visual-diff">
 			<d2l-navigation-button-close id="close"></d2l-navigation-button-close>
 		</div>
+		<div class="visual-diff">
+			<d2l-navigation-button-icon id="icon-text" icon="tier3:classes" text="Classes"></d2l-navigation-button-icon>
+		</div>
+		<div class="visual-diff">
+			<d2l-navigation-button-icon id="icon-text-disabled" icon="tier3:classes" text="Classes" disabled></d2l-navigation-button-icon>
+		</div>
+		<div class="visual-diff" id="icon-text-hidden-container" style="display: inline-block; padding: 0 30px 60px 30px;">
```
dlockhart, line 35: Padding out the parent so that the tooltip can fit nicely inside the space it creates.

d2l-navigation-button-icon.js
```diff
@@ -0,0 +1,112 @@
...
+	static get focusElementSelector() {
+		return 'button';
+	}
+
+	render() {
+		const ariaLabel = this.textHidden ? this.text : undefined;
+		const highlightBorder = !this.disabled ? html`<span class="d2l-navigation-highlight-border"></span>` : nothing;
+		const id = this.textHidden ? this._buttonId : undefined;
+		const text = !this.textHidden ? this.text : nothing;
+		const tooltip = this.textHidden ? html`<d2l-tooltip for="${this._buttonId}" for-type="label">${this.text}</d2l-tooltip>` : nothing;
```
eKoopmans, line 99: Nit: given that these are (mostly) grouped based on whether `textHidden` is true or false, it might make them more readable to group them, something like:

```js
_getTextSettings() {
  if (this.textHidden) return { ariaLabel: this.text, id: this._buttonId, text: nothing, tooltip: html`<d2l-tooltip for="${this._buttonId}" for-type="label">${this.text}</d2l-tooltip>` };
  return { ariaLabel: undefined, id: undefined, text: this.text, tooltip: nothing };
}

render() {
  const { ariaLabel, id, text, tooltip } = this._getTextSettings();
  const highlightBorder = this.disabled ? nothing : html`<span class="d2l-navigation-highlight-border"></span>`;
  // ...
}
```

eKoopmans (approved):

d2l-navigation-button-icon.js
dlockhart: Cool I like that -- updated!

svanherk (approved):

PR #154, opened 2022-07-11
fix: switch iterator to use button-icon

dlockhart:
Now that `<d2l-navigation-button-icon>` exists, the iterator component can use it directly instead of `<d2l-navigation-button>`. This required introducing an `icon-position` attribute so that the icon can be rendered on the right side for the "next" iterator button.

I also merged `<d2l-navigation-iterator-item>` into `<d2l-navigation-iterator>` since no one was using it directly and it wasn't adding any value. Deleted some tests as a result since they're now covered by the `d2l-navigation-button-icon` ones.

I'm expecting a bit of visual-diff change here since:
- The space between the icon and the next changed from `0.6rem` to `6px`
- Previously there was `1.2rem` of padding on the left and `1.5rem` on the right... the only explanation I can think of of this was it was intended to account for the size different between "Previous" and "Next"... but that only works in English. I've made them both `1.2rem`.
- The old disabled colours/opacity was inconsistent and still changed on-hover -- that's fixed now
- Previously a tooltip would always be shown on hover/focus, even if text was displayed. This was super redundant, and switching to the new button-icon component only shows tooltips when text is hidden.


components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -1,67 +1,62 @@
...
-					display: flex;
-					justify-content: space-between;
-					height: 3.3rem;
-					max-width: 20rem;
-				}
-				@media (max-width: 30.75rem) {
-					.d2l-navigation-iterator {
-						height: 3rem;
-					}
-				}
```
svanherk, line 50: Looks like previously the height of the iterator got smaller at a certain screen size?

dlockhart: Forgot to comment on this -- it definitely _looks_ that way, but there was nothing with a CSS class of `d2l-navigation-iterator`, so this rule wasn't doing anything.

[At one point](https://github.com/BrightspaceUILabs/navigation/commit/139fbfe9a9d7a327089610841e849d7c4226fe7c) there was a wrapper element with that class so this was doing something, but then [it was removed](https://github.com/BrightspaceUILabs/navigation/commit/05f13461bc69df18ef2e933d50162feca3e30887) without updating the responsive rule. So.... it's been like this for 3 years essentially since the component was introduced. At this point, I feel like it's probably safer to not re-introduce the responsiveness.

PR #156, opened 2022-07-11
fix: switch back link to use link icon

dlockhart:
A pretty simple swap here -- now that `<d2l-navigation-link-icon>` exists, `<d2l-navigation-link-back>` can just use it directly.

Had to fix this sequences usage, which was reaching into the shadow DOM:
https://github.com/BrightspaceHypermediaComponents/sequences/pull/442

I'm expecting a `1px` width increase as the spacing between the icon and text will go from `5px` to `6px`.


d2l-navigation-immersive.js
```diff
@@ -133,10 +133,6 @@ class D2LNavigationImmsersive extends DirMixin(PolymerElement) {
 				visibility: hidden;
 			}
 
-			d2l-navigation-link-back {
-				--d2l-navigation-link-back-left-padding: 12px;
```
dlockhart, line 137: This was the only place overriding this padding value -- I checked with Jeff and he saw no need to do this.

d2l-navigation-immersive.js
```diff
@@ -179,10 +175,6 @@ class D2LNavigationImmsersive extends DirMixin(PolymerElement) {
 					margin: 0 18px;
 					padding: 0 18px;
 				}
-				d2l-navigation-link-back {
-					--d2l-navigation-link-back-left-padding: 6px;
-					margin: 0
```
dlockhart, line 184: No margin was set, so this was doing nothing.

d2l-navigation-link-back.js
```diff
@@ -1,82 +1,39 @@
...
-	static get template() {
-		const template = html`
-			<style>
+
+	static get styles() {
+		return css`
 			:host {
 				display: inline-block;
 				height: 100%;
-				white-space: nowrap;
```
dlockhart, line 53: The `nowrap` is set in `d2l-navigation-link-icon`, so not needed.

d2l-navigation-link-back.js
```diff
@@ -1,82 +1,39 @@
...
-		if (text === undefined || text === null) {
-			return localize('back');
-		}
-		return text;
+	static get focusElementSelector() {
+		return 'd2l-navigation-link-icon';
 	}
+
+	render() {
+		const href = this.href ?  this.href : 'javascript:void(0);'; // backwards-compatible for uses before missing "href" threw exception
```
dlockhart, line 32: I was hoping to avoid this, but there are quite a few places (including the LMS) that are abusing it.

d2l-navigation-styles.js
```diff
@@ -30,6 +30,7 @@ export const highlightLinkStyles = css`
 		position: relative;
 		text-decoration: none;
 		vertical-align: middle;
+		white-space: nowrap;
```
dlockhart, line 33: This should have always been here, to avoid text/icon wrapping onto separate lines.

svanherk (approved):
Nice!

PR #162, opened 2022-07-13
fix: close tooltips when buttons are clicked

dlockhart:
Small fix that I think we want everywhere, but it also resolves this crazy issue in sequences:

tooltip-before

The issue there is that the target for the tooltip actually moves when the sidebar opens, and tooltip really has no mechanism to know that this happened and reposition itself. `ResizeObserver` and `MutationObserver` don't fire when things move, only when they change size or mutate. Either way, I still think just hiding the tooltip on-click is a nicer experience anyway.

After:
tooltip-after


svanherk (approved):
Nice!

dcostiuc (approved):

PR #174, opened 2022-07-25
fix: replace built-in padding with built-in gap in between iterator items

dlockhart:
For some reason, the iterator doesn't come with any spacing between the next/previous items or between the inner slot when it's in use. This causes things to get pretty tight in small spaces:

Screen Shot 2022-07-25 at 1 56 06 PM

Both consumers have realized this, and have introduced their own spacing. I'll be following up this change with the removal of the custom spacing:
- [LMS](https://cs.github.com/Brightspace/lms/blob/e8ff6ffb05dc163f83c69782e6b1a58232e7f8f3/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/ImmersiveNav/ImmersiveNav.css?q=%22d2l-navigation-iterator-middle%22#L37) (uses `1.2rem`)
- [consistent-eval](https://cs.github.com/Brightspace/consistent-evaluation/blob/edd391c10fbc812aa5265cc93e6f2786700b5cc8/components/header/consistent-evaluation-nav-bar.js?q=%22d2l-navigation-iterator%22#L197) (uses a value varying from `2.5rem` to `1rem` to nothing)

Additionally, the iterator comes with padding around the outside, but both places this is used inside immersive nav already come with padding. I'm removing this padding, allowing consumers to use whatever spacing they need.

The first change I'm making here is just to remove the `min-width` on the visual diff tests. This will highlight the actual problem. I'll merge those diffs in and then push the fix, which will show the change.


svanherk (approved):
Nice!

components/d2l-navigation-iterator/d2l-navigation-iterator.js
```diff
@@ -20,10 +20,10 @@ class NavigationIterator extends  LocalizeNavigationElement(LitElement) {
 			:host {
 				align-items: center;
 				display: flex;
+				gap: 1.2rem;
```
svanherk, line 23: Browser support looks good, there's a few versions of "legacy" that don't support `gap` this but the result would still be functional anyways

dlockhart: Yeah agreed, we've been using it a fair bit for similar things for a little while -- I'm good with where the support's at.

PR #179, opened 2022-07-27
Remove bad import of d2l-navigation-shared-styles

EvanLemieux:
Removing import of d2l-navigation-shared-styles that was causing issues with webpack in folio.


dlockhart (approved):
Thanks! This will get converted to Lit shortly as well.

PR #182, opened 2022-08-08
feat: convert band to Lit

dlockhart:
This migrates `<d2l-navigation-band>` to Lit. 🔥 

Not expecting any visual-diff changes here.

### DIFF
--- a/d2l-navigation-band.js
+++ b/d2l-navigation-band.js
@@ -26,10 +26,6 @@ class NavigationBand extends LitElement {
 				overflow-y: hidden;
 				scroll-behavior: smooth;
 			}
-			.d2l-navigation-scroll:before,
-			.d2l-navigation-scroll:after {
-				width: 2.439%;
-			}
 
 			.d2l-navigation-scroll[data-custom-scroll] {
 				/* Firefox Styles */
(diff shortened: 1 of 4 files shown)


d2l-navigation-band.js
```diff
@@ -1,62 +1,48 @@
...
+			.d2l-navigation-scroll[data-custom-scroll] {
 				/* Firefox Styles */
 				scrollbar-color: var(--d2l-color-galena) var(--d2l-color-sylvite);
 				scrollbar-width: thin;
-
-				/* IE Styles */
-				scrollbar-face-color: var(--d2l-color-galena);
-				scrollbar-arrow-color: var(--d2l-color-sylvite);
-				scrollbar-track-color: var(--d2l-color-sylvite);
-				scrollbar-shadow-color: var(--d2l-color-sylvite);
```
dlockhart, line 51: No longer required!

d2l-navigation-shared-styles.js
```diff
@@ -16,33 +16,18 @@ export const navigationSharedStyle = html`
 				position: relative;
 			}
 
-			.d2l-navigation-scroll:before,
-			.d2l-navigation-scroll:after {
-				width: 2.439%;
-			}
-
```
dlockhart, line 23: A couple places still using the "gutters" and "centerer" styles -- will delete this whole file once they're Lit.

d2l-navigation-band.js
```diff
@@ -1,85 +1,68 @@
...
+			.d2l-navigation-scroll::before,
+			.d2l-navigation-scroll::after {
 				content: '';
-				position: absolute;
 				height: 100%;
 				max-height: var(--d2l-navigation-band-slot-height, 1.5rem);
 				pointer-events: none;
+				position: absolute;
 				top: 0;
+				width: 2.439%;
```
dlockhart, line 56: This selector was defined twice and the linter caught it, so pulled this into here.

svanherk: Yeah, I think we had this in the shared file so that it was next to the gutters, so if their width was ever changed, this one would be too.  Could we store it as a variable to use in both places? 🤔 But not one exposed to consumers

svanherk: Or even a comment is probably fine

svanherk (approved):

PR #184, opened 2022-08-11
feat: add skip-nav support to the navbar component

dlockhart:
This is what our skip-nav looks like today, and no visual changes are happening as part of this work:

Screen Shot 2022-08-11 at 11 40 00 AM

As part of [DE49505](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fdefect%2F644089624677), we'd like to fix:
1. The semantics of our skip-nav: Make it a link instead of a button, since assistive technology users expect it to be a link.
2. Tweak the behaviour: Previously, we would focus on the page's `<h1>` and then fall back to a placeholder element after the navbar. In practice, the focusing on the placeholder never worked (focus would get lost). Instead, we first want to find a `<main>` element (or `[role="main"]`), then fall back to the `<h1>` and finally fall back to the first focusable element after the navbar.

At first I just added this functionality directly to `<d2l-navigation>`, but because that made it extend `LocalizeDynamicMixin`, I noticed that the extra rendering delay that caused broke a bunch of timing-sensitive things in the LMS (the link chomping specifically). So to be extra safe, I instead made `<d2l-navigation-skip>` its own component.


d2l-navigation-skip.js
```diff
@@ -0,0 +1,70 @@
+import { css, html, LitElement } from 'lit';
+import { LocalizeNavigationElement } from './components/localize-navigation-element.js';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { querySelectorComposed } from '@brightspace-ui/core/helpers/dom.js';
+
+class NavigationSkip extends LocalizeNavigationElement(RtlMixin(LitElement)) {
+
+	static get styles() {
+		return css`
```
dlockhart, line 9: This style is [copied from here](https://search.d2l.dev/xref/lms/platformTools/navbars/D2L.PlatformTools.Navbars/Web/Desktop/Rendering/Views/Css/SkipNav.css?r=07c221d1) and tweaked to pass linting.

d2l-navigation-skip.js
```diff
@@ -0,0 +1,70 @@
...
+				right: 25%;
+			}
+		`;
+	}
+
+	render() {
+		return html`<a tabindex="0" @click="${this._handleSkipNav}" @keydown="${this._handleKeyDown}">${this.localize('skipNav')}</a>`;
+	}
+
+	_handleKeyDown(e) {
```
dlockhart, line 49: Since the link doesn't have an `href`, we need to give it a `tabindex` and handle the ENTER key.

d2l-navigation-skip.js
```diff
@@ -0,0 +1,70 @@
...
+
+	_handleSkipNav() {
+		const elem = querySelectorComposed(document, 'main') ||
+			querySelectorComposed(document, '[role="main"]') ||
+			querySelectorComposed(document, 'h1');
+		if (elem) {
+			elem.tabIndex = -1;
+			elem.focus();
+		} else {
+			this.dispatchEvent(new CustomEvent('d2l-navigation-skip-fail', { bubbles: false, composed: false }));
```
dlockhart, line 64: Couldn't find something to focus on, so we fire this event and defer to the parent navbar to "do the right thing". That's because skip-nav doesn't really know where the navbar begins/ends.

lang/ar.js
```diff
@@ -3,5 +3,6 @@
 export default {
 	"back": "العودة",
 	"next": "التالي",
-	"previous": "السابق"
+	"previous": "السابق",
+	"skipNav": "تجاوز إلى المحتوى الرئيسي"
```
dlockhart, line 7: These lang terms were copied from the LMS -- I'll remove them from the LMS as part of a PR there to adopt this.

d2l-navigation-skip.js
```diff
@@ -0,0 +1,70 @@
...
+		return css`
+			a {
+				left: -1000px;
+				overflow: hidden;
+				position: absolute;
+				width: 1px;
+			}
+			:host([dir="rtl"]) a {
+				left: auto;
+				right: -10000px;
```
svanherk, line 18: Lol do you think there's a reason we had left as one thousand and right as ten thousand? 😝

d2l-navigation-skip.js
```diff
@@ -0,0 +1,70 @@
...
+			}
+			a:active,
+			a:focus {
```
svanherk, line 21: I'm assuming the styles left out here were the default (like `box-sizing`) or weren't doing anything (like `vertical-align`)?

d2l-navigation-skip.js
dlockhart: Haha good eye! Likely a mistake -- I've made them both `10,000`, which matches what our [offscreen styles](https://github.com/BrightspaceUI/core/blob/main/components/offscreen/offscreen.js#L7) use.

d2l-navigation-skip.js
dlockhart: Sorry, I meant to comment on that -- those attributes along with a few others were button-specific and no longer needed now that it's a link. Based on [the original change from link to button](https://old-git.d2l.dev/projects/CORE/repos/platformtools/commits/1c8499b4a9c6aa64ff191a56254398d55b657f6f#navbars/D2L.PlatformTools.Navbars/Web/Desktop/Rendering/Views/Css/SkipNav.css).

svanherk: Ah gotcha, makes sense!

d2l-navigation.js
```diff
@@ -26,14 +34,27 @@ class Navigation extends LitElement {
...
-			<d2l-navigation-band><slot name="navigation-band"></slot></d2l-navigation-band>
+			${skipNav}<d2l-navigation-band><slot name="navigation-band"></slot></d2l-navigation-band>
 			<slot></slot>
 			<div class="d2l-navigation-shadow-drop-border"></div>
 		`;
 	}
 
+	_handleSkipNavFail() {
+		const nextFocusable = getNextFocusable(this.shadowRoot.querySelector('.d2l-navigation-shadow-drop-border'));
+		if (nextFocusable) {
```
svanherk, line 53: What if there is no focusable element on the page?  I'm imagining a case where its all paragraphs, or maybe in an `iframe` (will nextFocusable go into an `iframe`?)  I thought before we created an element for this case, but sounds like from the PR description this wasn't working?

d2l-navigation-skip.js
```diff
@@ -0,0 +1,70 @@
...
+			this._handleSkipNav();
+		}
+	}
+
+	_handleSkipNav() {
+		const elem = querySelectorComposed(document, 'main') ||
+			querySelectorComposed(document, '[role="main"]') ||
+			querySelectorComposed(document, 'h1');
+		if (elem) {
+			elem.tabIndex = -1;
```
eKoopmans, line 61: What's the purpose of removing the tabIndex of the main/h1 element? Is this maybe supposed to be `this.tabIndex = -1`?

d2l-navigation.js
dlockhart: Yeah so in the case of no `<main>`, no `<h1>` and no focusable element, nothing will happen. Correct, we did have a [placeholder span](https://search.d2l.dev/xref/lms/platformTools/navbars/D2L.PlatformTools.Navbars/Web/Desktop/Rendering/Views/SkipNavView.Renderer.cs?r=e8af20fb#50), but in my testing it would move focus there but then the next tab always went back to the top of the page.

The `iframe` case is interesting -- I tested it out with an iFRA. Our fullscreen FRA pages still render a `<div role="main">`, but if I remove that the way `getNextFocusable` is written, it will still move focus to the `iframe` itself, but not go inside the iframe since that's cross-origin.

d2l-navigation-skip.js
dlockhart: This isn't so much removing it, as setting it to `-1`, which is what allows us to call `focus()` on an otherwise non-focusable element. Both `h1` and `main` are not inherently focusable, i.e. they don't get keyboard stops as you tab around. But `-1` lets us programatically focus on them without introducing a keyboard stop, which a `0` value would do.

eKoopmans: Right, my mistake. Is there any risk that the element already has an intentional tabIndex that's not `-1`?

d2l-navigation.js
svanherk: Ah nice, yeah looks like we consider `iframe`s focusable: https://github.com/BrightspaceUI/core/blob/main/helpers/focus.js#L8

d2l-navigation-skip.js
dlockhart: I suppose there's some risk, although the [old code was has been this for 3 years](https://search.d2l.dev/xref/lms/platformTools/navbars/D2L.PlatformTools.Navbars/Web/Desktop/Rendering/Views/Daylight/Navigation.ClassInitializer.js?r=af44aa93#16) with the `h1`. I'm not worried about it.

svanherk (approved):
Looks good to me!  I'm a tad worried that focusable `span` was working in some weird client-specific edge case, but if so hopefully that case is covered now by the additional `main` targeting

dlockhart:
> I'm a tad worried that focusable span was working in some weird client-specific edge case

We render a `<div role="main">` on all our MVC and legacy pages (even blank page), so I think we're good.

PR #186, opened 2022-08-15
feat: add ability to hide the highlight border

dlockhart:
After discussing [this use case in consistent-eval](https://github.com/Brightspace/consistent-evaluation/blob/main/components/footer/consistent-evaluation-footer.js#L130-L139), Jeff and I came to the conclusion that the best outcome would be to just continue to support this.

Screen Shot 2022-08-10 at 11 46 25 AM

I'm not adding this to `<d2l-navigation-link-icon>` however in order to limit the surface area for now.


svanherk (approved):
Looks good!

PR #188, opened 2022-08-19
Fix d2l-navigation-iterator import

ChrisLabattD2L:
I've added how to `npm install` this repo since it was missing and also corrected the import for the `d2l-navigation-iterator` component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,12 @@
 
 A series of web components for top level navigation use in D2L applications.
 
+## Installation
+
+```shell
+npm install @brightspace-ui-labs/navigation
+```
+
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Thanks! I'm 90% through converting this to Lit and was going to do a full README audit at that point, but this is still great.

ChrisLabattD2L:
> Thanks! I'm 90% through converting this to Lit and was going to do a full README audit at that point, but this is still great.

Didn't realize that conversion was underway! I was just trying to get it set up and figured I'd just add it since it was missing 😄

PR #192, opened 2022-10-21
Update d2l-demo-snippet fullscreen to full-width

bearfriend:
The `d2l-demo-snippet` attribute `fullscreen` is being renamed to full-width. These changes will need to happen more or less at the same time as core is updated.

Once the changes are released in `core` this PR will be updated to bump the dependency version.

BrightspaceUI/core#2926

[US144714](https://rally1.rallydev.com/#/?detail=/userstory/665191647783&fdp=true): Add fullscreen switch to d2l-demo-snippet

### DIFF
--- a/demo/navigation.html
+++ b/demo/navigation.html
@@ -40,7 +40,7 @@
 	<body>
 		<d2l-demo-page page-title="d2l-navigation">
 			<h2>Header and Footer</h2>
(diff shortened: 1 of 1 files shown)


bearfriend:
There is no `package-lock` and `package.json` targets `^2`. This should be ready to merge.

svanherk (approved):

PR #196, opened 2023-03-14
use focusWithKeyboard instead of forceFocusVisible

dlockhart:


test/navigation.visual-diff.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-invalid-this */
```
dlockhart, line 1: Not sure what was going on with these ignore rules, but no longer necessary.

eKoopmans (approved):
Looks good, thanks!

PR #197, opened 2023-06-12
DE51509 - add tooltip-position property to d2l-navigation-button-icon

derekmoran:
[DE51509](https://rally1.rallydev.com/#/42960320374ud/custom/236803223728?detail=%2Fdefect%2F675164753105): Tool tip for iterating between students in CE is cut off

Pairs with this PR:
* https://github.com/Brightspace/consistent-evaluation/pull/1631

image

PR #199, opened 2023-06-12
update visual diffs

dlockhart:
So this can happen before #197.


test/immersive.visual-diff.js
```diff
@@ -76,7 +76,7 @@ describe('d2l-navigation-immersive', () => {
 
 		it('back-button', async function() {
 			await initTest(page, 'normal');
-			await focusWithKeyboard(page, 'd2l-navigation-immersive');
+			await focusWithKeyboard(page, ['d2l-navigation-immersive', 'd2l-navigation-link-back']);
```
dlockhart, line 79: Not sure how this was ever working before.

test/link.visual-diff.html
```diff
@@ -16,8 +16,8 @@
 	</head>
 	<body class="d2l-typography">
 		<div class="visual-diff">
-			<div style="display: inline-block;" id="base">
-				<d2l-navigation-link href="https://www.d2l.com" text="D2L">D2L</d2l-navigation-link>
+			<div style="display: inline-block;" id="base-container">
+				<d2l-navigation-link id="base" href="https://www.d2l.com" text="D2L">D2L</d2l-navigation-link>
```
dlockhart, line 20: Same here. It was focusing on the outer `<div>`, which shouldn't have done anything.

test/immersive.visual-diff.js
dlockhart: This stuff definitely broke [with these changes](https://github.com/BrightspaceUILabs/navigation/pull/196) since our old focus method recursively searched for things to focus on... but I'm not sure why CI didn't catch the problems.

svanherk:
Yeah that's really weird... could it have somehow been working because of the old "tab" logic?  This PR was merged after these changes were made here: https://github.com/BrightspaceUI/visual-diff/pull/96

svanherk (approved):

derekmoran (approved):

dlockhart:
> Could it have somehow been working because of the old "tab" logic?

That has to be it. It just so happens to be the first thing on the page, so that old logic would have "tabbed" right to it! So excited for isolated fixtures...

PR #201, opened 2023-10-05
fix: switch to non pseudo-elements to resolve Firefox tabindex

dlockhart:
[DE54851](https://rally1.rallydev.com/#/15545167705d/custom/21568985922?detail=%2Fdefect%2F721029304029).

This is a weird one. When the navigation band has slotted content (like site tabs), we use `::before` and `::after` pseudo elements positioned absolutely to display a faded edge when scrolled. Unfortunately, Firefox interprets the `::after` one as "overflowing" its container and whenever Firefox detects a scrollable panel, it gives it a tab stop. The end result is that all navigation bands in Firefox get a tab stop, even if they're empty!

The fix here was a result of trial and error. Replacing them with `<div>`s before and after the slot wasn't enough -- I had to move the "after" one to be before the gutters wrapper. But because they're absolutely positioned, their DOM order doesn't matter so this fix worked.


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         run: |
           npm install
           npm install @web/test-runner-playwright --no-save
-          npx playwright install-deps
+          npx playwright install --with-deps
```
dlockhart, line 17: Since I'm missing it after 1 day, I might convert this to the new test runner & vdiff in a follow-up PR.

dbatiste (approved):

svanherk (approved):
Crazy!

PR #202, opened 2023-10-05
Migrate to d2l-test-runner

dlockhart:
Modernizes this repo a bit, including linting and switch to d2l-test-runner and new vdiffs.


test/link.vdiff.js
```diff
@@ -0,0 +1,61 @@
...
+			text="D2L"></d2l-navigation-link-image>
```
dlockhart, line 14: The relative path to the image was failing in CI, so ended up just converting it to base64 and using a data URI.

svanherk:
I think you want to put the Required check on `build (pull_request)`, not `build results (pull_request)`.  It mostly won't matter because they'll match, but in the dependabot case we won't have the "results" one

svanherk (approved):

PR #213, opened 2024-06-11
feat: Add allow-overflow to navigation-immersive

eKoopmans:
We're adding a dropdown to our folio immersive-nav-bar, and the `overflow: hidden` was causing issues. I don't know if any consumers rely on the overflow being hidden, so I added a new attribute to disable it.

Here's what the demo looks like without `allow-overflow`:

image

And with:

image


d2l-navigation-immersive.js
```diff
@@ -19,6 +19,10 @@ class D2LNavigationImmsersive extends DirMixin(PolymerElement) {
 
 	static get properties() {
 		return {
+			allowOverflow: {
+				type: Boolean,
+				reflectToAttribute: true
+			},
```
dlockhart, line 25: @dbatiste is working on some changes to how our dropdowns work... not sure if it might actually solve this?

dbatiste: Yeah it should. We're keeping it sorta under wraps for a little bit while Margaree kicks the tires.

Looks like this was added back here... https://github.com/BrightspaceUILabs/navigation/commit/3ef8acdc455124d521b523a49661324e1f03c40c

eKoopmans: Ooh exciting, popup API stuff?

dlockhart: Dave, does it make sense to add a story/task to your stuff to circle back to this change when everything is finished and undo it?

eKoopmans: Yeah I noticed it was from the original PR that added `navigation-immersive`, not sure what the motivation was for the overflow specifically. The regular (non-immersive) navigation allows overflow, which is how things like this work:

image

dbatiste: @dlockhart , Yeah I think so.

I suspect this is to handle this case. But perhaps not just that case.
image

dbatiste: Yeah, I remember we ran into this issue pretty much immediately, when we implemented the dropdowns and added them to the main nav. I recall having to find a way to avoid the overflow.

dlockhart (approved):

d2l-navigation-immersive.js
eKoopmans: Just confirmed that we're already separately handling the overflow from Dave's screenshot, glad you called it out though:

image

dbatiste: > Ooh exciting, popup API stuff?

Eventually, hopefully/yes. But initially, it will just be adapting the existing positioning code to unblock some other things.

dbatiste (approved):

eKoopmans:
Thanks for the speedy review Daves!

PR #216, opened 2024-06-17
Migrate some nav elements to Lit

dlockhart:
This migrates the header, footer and "immersive" elements from Polymer to Lit. I needed to make an update to the immersive element, so figured this was worth it.


bearfriend (approved):
🎉

dbatiste (approved):

PR #219, opened 2024-06-17
feat: add "back-link-text-short" attribute

dlockhart:
Instead of migrating its visual-diff tests, I'd like to Graveyard the [immersive-nav Labs repo](https://github.com/BrightspaceUILabs/immersive-nav/tree/main) and switch its 3 usages over to the immersive navigation component in this repo.

The one feature gap is that the other repo supports providing a shorter "back link text" which gets rendered at smaller viewports. This adds support for that feature here.


dbatiste (approved):
LGTM!

PR #221, opened 2024-06-17
GAUD-6565: add back link click event

dlockhart:
Two of the Insights usages ([example](https://github.com/Brightspace/insights-adoption-dashboard/blob/a5904fc9881fbc12bb65f97bbe1c1a8863afe2cb/insights-adoption-dashboard.js#L142)) are wiring up to a click event handler on the back link to apply some extra logic. This adds an event for them to do that.


dbatiste (approved):
LGTM!

